PR #2, opened 2017-06-08
Replace unittest2 with unittest

crossroads1112:
Per `unittest2`'s [description](https://pypi.python.org/pypi/unittest2), and Python2.7's [documentation](https://docs.python.org/2.7/library/unittest.html):

> unittest2 is a backport of the new features added to the unittest testing framework in Python 2.7 and onwards

Is there a reason `check50` needs to be compatible with pre-2.7 Python (especially since its written in Python 3)?

### DIFF
--- a/check50.py
+++ b/check50.py
@@ -11,7 +11,7 @@ import shutil
 import sys
 import tempfile
 import traceback
-import unittest2
+import unittest
 
 from functools import wraps
 from termcolor import cprint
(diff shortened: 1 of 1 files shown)


dmalan:
Good catch, we can assume >= 2.7, since SQLAlchemy requires the same, and Mac OS comes with 2.7. It'll ultimately be written in Python 2, though, not 3.

crossroads1112:
Sorry, I think I'm misunderstanding. If we can assume at least 2.7, why use `unittest2`? unittest2 == unittest for every Python version >= 2.7

dmalan:
Right, no need, already closed PR!

PR #4, opened 2017-06-15
pset 2 checks

hljames:


checks/caesar.py
```diff
@@ -0,0 +1,47 @@
+import os
+import sys
+
+sys.path.append(os.getcwd())
```
dmalan: We should avoid needing this in individual checks. This should be pulled into parent class if need be.

checks/caesar.py
```diff
@@ -0,0 +1,47 @@
...
+from check50 import File, Test, Error, check
+
+class Vigenere(Test):
```
dmalan: @brianyu28, to avoid new nomenclature, should we perhaps just call our class `check50.TestCase`? Or something like `Suite`, as in a "suite of checks"?

checks/caesar.py
```diff
@@ -0,0 +1,47 @@
...
+        """caesar.c exists."""
+        super().exists("caesar.c")
+
+    @check("exists")
+    def compiles(self):
+        """caesar.c compiles."""
+        self.spawn("clang -o caesar caesar.c -lcs50").exit(0)
+
+    @check("compiles")
+    def test_1a(self):
```
dmalan: No need for `test_` convention. Confer with @brianyu28 on how best to name methods moving forward.

checks/initials/more.py
```diff
@@ -0,0 +1,47 @@
...
+        """Outputs HLJ for hailey lynn james"""
+        self.spawn("./initials").stdin("hailey lynn james", prompt=False).stdout("HLJ\n", "HLJ").exit(0)
+
+    @check("compiles")
+    def test_haileyJames(self):
+        """Outputs HJ for hailey James"""
+        self.spawn("./initials").stdin("hailey James", prompt=False).stdout("HJ\n", "HJ").exit(0)
+
+    @check("compiles")
+    def test_hailey--James(self):
```
dmalan: Does this work? Avoid hyphens in method names.

checks/caesar.py
brianyu28: Probably best not to use `Suite` because the `TestSuite` class in `unittest2` already exists and serves a different function. Could rename `Test` to `TestCase` though if that's preferable.

brianyu28: Also, @hljames, this should probably be `class Caesar` instead of `class Vigenere`?

dmalan: Sure, @brianyu28, for now, why don't we just mimic `unittest`'s own class names, since we're just tweaking them for our purposes. Seems easier for new readers (myself included) not to introduce new semantics.

brianyu28: :+1:

checks/initials/more.py
```diff
@@ -0,0 +1,47 @@
...
+        """Outputs HJ for hailey       James"""
+        self.spawn("./initials").stdin("hailey       James", prompt=False).stdout("HJ\n", "HJ").exit(0)
+
+    @check("compiles")
+    def test_--haileyJames--(self):
```
dmalan: Dashes?

checks/caesar.py
crossroads1112: This can't be put into a parent class because it is what allows us to access the parent class. 

Once check50 is installed as a package locally, this won't be necessary

dmalan: 👍

checks/initials/more.py
crossroads1112: `checks/{initials/*, *.py}` should not actually be in the PR at all. The actual tests are in `checks/cs50/2017/x/{initials, caesar, vigenere}`

crossroads1112:
@dmalan Updated to latest API

PR #5, opened 2017-06-16
Update: Added .gitignore

ankitrgadiya:
I added .gitignore to automatically ignore `__pychache__` from the next commit.


brianyu28:
Good catch, @ankitrgadiya! Actually just fixed this in the root `.gitignore` per https://github.com/cs50/check50/blob/9cada53f42a75a1a3f660846c6cf254106892665/.gitignore#L2, but thank you!

PR #8, opened 2017-06-19
checks for pset3 through compilation and existence of fifteen -- inco…

hljames:
…mplete

### DIFF
--- a/checks/cs50/2017/x/fifteen/checks.py
+++ b/checks/cs50/2017/x/fifteen/checks.py
@@ -1,0 +1,141 @@
+from check50 import *
+
+
+class Fifteen(Checks):
+
+    @check()
+    def exists(self):
+        """fifteen.c exists."""
+        super(Fifteen, self).exists("fifteen.c")
+
+    @check("exists")
+    def compiles(self):
+        """fifteen.c compiles."""
+        # Remove sleeps from student code
+        self.spawn("sed -i='' '/#include <unistd.h>/a \\\n#define usleep(x)' fifteen.c").exit()
+
+        cflags = "-std=c99 -Wall -Werror -Wno-deprecated-declarations"
+        # Replace student's draw with our own
+        self.include("simple_draw.c")
+        self.spawn("clang -S {} fifteen.c -o fifteen.S".format(cflags)).exit(0)
+        self.replace_fn("draw", "simple_draw", "fifteen.S")
+        self.spawn("clang {} -o fifteen simple_draw.c fifteen.S -lcs50".format(cflags)).exit(0)
+
+    @check("compiles")
+    def init3(self):
+        """3x3 board: init initializes board correctly"""
+        self.spawn("./fifteen 3").stdout("8-7-6|5-4-3|2-1-0\n").stdout("Tile to move:")
+
+    @check("init3")
+    def invalid8(self):
+        """3x3 board: catches moving 8 an illegal move"""
+        self.spawn("./fifteen 3").stdin("8")                    \
+                                 .stdout("Illegal move.")       \
+                                 .stdout("8-7-6|5-4-3|2-1-0\n") \
+                                 .stdout("Tile to move:")
+    @check("init3")
+    def valid1(self):
+        """3x3 board: catches moving 1 as a legal move"""
+        self.spawn("./fifteen 3").stdin("1")                    \
+                                 .stdout("8-7-6|5-4-3|2-0-1\n") \
+                                 .stdout("Tile to move:")
+
+    @check("init3")
+    def move_up2(self):
+        """3x3 board: move blank up twice"""
+        self.spawn("./fifteen 3").stdin("3")                    \
+                                 .stdout("8-7-6|5-4-0|2-1-3\n") \
+                                 .stdin("6")                    \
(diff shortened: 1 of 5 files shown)


checks/caesar.py
```diff
@@ -0,0 +1,47 @@
+import os
+import sys
+
+sys.path.append(os.getcwd())
+from check50 import File, Test, Error, check
```
dmalan: @brianyu28 can we simplify this to `from check50 import *`, so that authors don't need to remember what to import? Or will that pull in too much?

checks/fifteen.py
```diff
@@ -0,0 +1,68 @@
+import os
+import sys
+
+sys.path.append(os.getcwd())
+from check50 import File, Test, Error, check
+
+class Fifteen(Test):
```
dmalan: @hljames, @brianyu28 and I discussed how we could handle pset3 (by swapping out students' `draw` function as needed). Do just discuss with him.

checks/initials/less.py
```diff
@@ -0,0 +1,37 @@
...
+        """Outputs HLJ for hailey lynn james"""
+        self.spawn("./initials").stdin("hailey lynn james", prompt=False).stdout("HLJ\n", "HLJ").exit(0)
+
+    @check("compiles")
+    def test_haileyJames(self):
+        """Outputs HJ for hailey James"""
+        self.spawn("./initials").stdin("hailey James", prompt=False).stdout("HJ\n", "HJ").exit(0)
+
+    @check("compiles")
+    def test_BRIAN`(self):
```
dmalan: What's this backtick?

checks/fifteen.py
brianyu28: @crossroads1112 worked out a way to do this today by swapping out instances of `callq foo` in the assembly file to `callq bar`, where `foo` is the student's version of the function and `bar` is ours.

checks/caesar.py
brianyu28: Sorry for missing this-- we could definitely simplify to `from check50 import *` without issue if that's easier.

dmalan: Let's start with `*` then, to keep things marginally simpler.

checks/fifteen.py
```diff
@@ -0,0 +1,139 @@
...
+    def invalid_center(self):
+        """3x3 board, move blank left (tile 1) then up (tile 4), then try to move tiles 1, 2, 6, 8"""
+        child = self.spawn("./fifteen 3")
+        for move in ["2", "4"]:
+            child.stdin(move)
+            try:
+                child.child.expect('Tile to move:', timeout=3)
+            except pexpect.TIMEOUT:
+                raise Error()
+            for move in ["5", "6", "7", "8"]:
```
dmalan: Should we not be waiting for subsequent `Tile to move: ` outputs?

checks/initials/more.py
```diff
@@ -0,0 +1,47 @@
...
+        """Outputs HLJ for hailey lynn james"""
+        self.spawn("./initials").stdin("hailey lynn james", prompt=False).stdout("HLJ\n", "HLJ").exit(0)
+
+    @check("compiles")
+    def test_haileyJames(self):
+        """Outputs HJ for hailey James"""
+        self.spawn("./initials").stdin("hailey James", prompt=False).stdout("HJ\n", "HJ").exit(0)
+
+    @check("compiles")
+    def test_hailey--James(self):
```
dmalan: Why double dashes?

crossroads1112:
@dmalan Updated with commits squashed

checks/cs50/2017/x/fifteen/checks.py
```diff
@@ -0,0 +1,152 @@
...
+    @check()
+    def exists(self):
+        """fifteen.c exists."""
+        super().exists("fifteen.c")
+
+    @check("exists")
+    def compiles(self):
+        """fifteen.c compiles."""
+        # Remove sleeps from student code
+        self.spawn("sed -i='' '/#include <unistd.h>/a \\\n#define usleep(x)' fifteen.c").exit(0)
```
dmalan: @brianyu28 if a line like this fails, will student be confused, since we'll log that we expected an exit code of 0?

checks/cs50/2017/x/fifteen/checks.py
```diff
@@ -0,0 +1,152 @@
...
+
+        cflags = "-std=gnu99 -Wall -Werror -Wno-deprecated-declarations"
+        # Replace student's draw with our own
+        self.include("simple_draw.c")
+        self.spawn("clang -S {} fifteen.c -o fifteen.S".format(cflags)).exit(0)
+        self.replace_fn("draw", "simple_draw", "fifteen.S")
+        self.spawn("clang {} -o fifteen simple_draw.c fifteen.S -lcs50".format(cflags)).exit(0)
+
+    # probably don't need this check -- we can just test the students work based on moves
```
dmalan: Should this be removed then?

checks/cs50/2017/x/fifteen/checks.py
crossroads1112: We can fix this by just passing no argument to `exit`

PR #9, opened 2017-06-19
partial pset 4

hljames:


dmalan:
Next time, do just try to submit only the pset4 stuff in a pset4 PR!

dmalan:
PS, closing for now, but aok to open new one once pset4's further along.

PR #14, opened 2017-06-20
Draft pset7 checks

brianyu28:
This is an initial draft of C$50 Finance checks using Selenium with `check50 v2`. I looked into other web testing alternatives, but honestly Selenium seems to be the best one out there. The key techniques employed in this test module are:

- replacing the student's `lookup()` function with a deterministic one that lets us test purchases and sales more effectively, as implemented in `override_lookup`.
- using PhantomJS with Selenium to allow for headless browser testing.
- making some assumptions about the way the student implemented the problem: namely requiring that input fields have a deterministic `name` attribute, and that buttons have deterministic text.

Thus far the tests seem to be working pretty well, although admittedly they're a little bit slow. (Each check takes about 5-6 seconds, for a total of 75 seconds or so runtime, details below). This is in part because, to maintain our system of copying files into a new directory for each check, we need to re-start the web server each time and close it each time.

```
$ time python ../check50.py finance
:) application.py exists.
:) application starts up.
:) user can register for account.
:) registration with an empty field fails.
:) registration with password mismatch fails. To 
:) registration rejects duplicate username.
:) quote handles invalid ticker symbol.
:) quote handles blank ticker symbol.
:) quote handles valid ticker symbol.
:) buy handles invalid ticker symbol.
:) buy handles fractional, negative, and non-numeric shares.
:) buy handles valid purchase.
:) sell handles invalid ticker symbol.
:) sell handles valid sell.
:) history shows transactions.

real	1m15.986s
user	0m49.448s
sys	0m8.179s
```

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -1,2 +1,3 @@
 *.c
 __pycache__/
+pset7/*
--- a/checks/finance.py
+++ b/checks/finance.py
@@ -1,13 +1,305 @@
 import os
 import re
+import requests
 import sys
 
 sys.path.append(os.getcwd())
-from check50 import TestCase, Error, check
+from check50 import File, TestCase, Error, check
+from selenium import webdriver
 
 class Finance(TestCase):
+
+    BASE_PATH = "http://127.0.0.1:5000"
     
+    def server_up(self):
+        env = {
+            "FLASK_APP": "application.py"
+        }
+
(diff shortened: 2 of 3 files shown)


brianyu28:
Closing to revise to new file structure, will submit another PR

PR #17, opened 2017-06-22
Pset6

hljames:
python porting checks


dmalan:
@hljames, mind PRing just pset6?

PR #18, opened 2017-06-22
Add pset7 checks

brianyu28:


checks/cs50/2017/x/finance/checks.py
```diff
@@ -1,13 +1,305 @@
 import os
 import re
+import requests
 import sys
 
 sys.path.append(os.getcwd())
-from check50 import TestCase, Error, check
+from check50 import File, TestCase, Error, check
```
dmalan: Do we want to promote this convention among staff or just `*`?

checks/cs50/2017/x/finance/checks.py
```diff
@@ -1,13 +1,305 @@
...
+from selenium import webdriver
 
 class Finance(TestCase):
+
+    BASE_PATH = "http://127.0.0.1:5000"
```
dmalan: For local testing in Docker, would `0.0.0.0` be more robust, since tends to bind to an internal IP? And `8080` for consistency with open IDE ports (and CS50 CLI's)?

checks/cs50/2017/x/finance/checks.py
```diff
@@ -1,13 +1,305 @@
...
+    
+    def server_up(self):
+        env = {
+            "FLASK_APP": "application.py"
+        }
+
+        self.server = self.spawn("flask run", env=env)
+        self.driver = webdriver.PhantomJS()
```
dmalan: Feels likes this `driver` should be assigned by `check50` library?

checks/cs50/2017/x/finance/checks.py
```diff
@@ -1,13 +1,305 @@
...
+
+    def server_down(self):
+        self.driver.close()
+        self.server.kill()
```
dmalan: Should library also have an `excepthandler` that calls this too?

checks/cs50/2017/x/finance/checks.py
```diff
@@ -1,13 +1,305 @@
...
+
+    def override_lookup(self):
+        """Replaces student's lookup function with a deterministic one."""
+        self.append_code("helpers.py", File("lookup.py"))
```
dmalan: @crossroads1112, what's the equivalent C feature look like as of now?

checks/cs50/2017/x/finance/checks.py
```diff
@@ -1,13 +1,305 @@
...
+    
+    def register(self, username, password, confirmation, outcome):
+        """Checks that registering results in success or failure."""
+
+        # go to register page
+        self.driver.get("{}/register".format(Finance.BASE_PATH))
```
dmalan: `BASE_PATH` or, really, `.get` should probably become part of check50's library, perhaps with just a `get` call that's hard-coded for `http://0.0.0.0:8080`?

checks/cs50/2017/x/finance/checks.py
```diff
@@ -1,13 +1,305 @@
...
+                fields_filled += 1
+                field.send_keys(username)
+            elif name == "password":
+                fields_filled += 1
+                field.send_keys(password)
+            elif name == "confirmation":
+                fields_filled += 1
+                field.send_keys(confirmation)
+        if fields_filled != 3:
+            raise Error("Could not fill in username and/or password.")
```
dmalan: For actual fall check, probably better if feedback is more granular, such that we report which param(s) are missing?

checks/cs50/2017/x/finance/checks.py
```diff
@@ -1,13 +1,305 @@
...
+    
+        # search for register button
+        buttons = self.driver.find_elements_by_tag_name("button")
+        register_button = None
+        for button in buttons:
+            if button.text == "Register":
+                register_button = button
+        if register_button == None:
+            raise Error("Could not find register button.")
```
dmalan: Like this, but for fields.

checks/cs50/2017/x/finance/checks.py
```diff
@@ -1,13 +1,305 @@
...
+        register_button.click()
+
+        if "Apology" in self.driver.title and outcome:
+            raise Error("Expected registration success, not rejection.")
+        elif "Apology" not in self.driver.title and not outcome:
```
dmalan: Feels like we should also be abstracting away `driver`, maybe calling it `document` or such?

checks/cs50/2017/x/finance/checks.py
```diff
@@ -1,13 +1,305 @@
...
+        login_button.click()
+
+        if "Apology" in self.driver.title:
+            raise Error("Expected login to succeed, not fail.")
+
+    def quote(self, ticker, outcome, value=None):
+        """Checks that getting a quote results in desired outcome."""
+
+        # go to quote page
+        self.driver.get("{}/quote".format(Finance.BASE_PATH))
```
dmalan: To be clear, this call could become, e.g., `self.get("/quote")`.

checks/cs50/2017/x/finance/checks.py
```diff
@@ -1,13 +1,305 @@
...
+        buttons = self.driver.find_elements_by_tag_name("button")
+        quote_button = None
+        for button in buttons:
+            if button.text == "Quote":
+                quote_button = button
+        if quote_button == None:
+            raise Error("Could not find quote button.")
+        quote_button.click()
+
+        if "Apology" in self.driver.title and outcome:
```
dmalan: Is there perhaps a more robust way to check for these "errors"? Most sites probably spit out 200s anyway, but we could tweak `apologize` to spit out, e.g., 400?

checks/cs50/2017/x/finance/checks.py
```diff
@@ -1,13 +1,305 @@
...
+            raise Error("Expected quote success, not failure.")
+        elif "Apology" not in self.driver.title and not outcome:
+            raise Error("Expected quote failure, not success.")
+        elif outcome and value:
+            # check the output against the given value
+            if str(value) not in self.driver.find_element_by_tag_name("body").text:
+                raise Error("Expected quote not found.")
+
+    def transaction(self, transaction, symbol, shares, outcome, value=None, remaining=None):
```
dmalan: Should we adopt a convention of `_`-prefixed methods for "private" methods? Or think the lack of `@check` suffices?

checks/cs50/2017/x/finance/checks.py
```diff
@@ -1,13 +1,305 @@
...
+                fields_filled += 1
+                field.send_keys(symbol)
+            elif name == "shares":
+                fields_filled += 1
+                field.send_keys(shares)
+        if fields_filled != 2:
+            raise Error("Could not fill in symbol and/or shares.")
+
+        # search for buy button
+        buttons = self.driver.find_elements_by_tag_name("button")
```
dmalan: Should we wrap these most commonly used `driver` methods (this one, get, post, etc.) so that we can add automatic logging to explain to students what's going on?

checks/cs50/2017/x/finance/checks.py
```diff
@@ -1,13 +1,305 @@
...
+        buy_button = None
+        for button in buttons:
+            if button.text == transaction.capitalize():
+                buy_button = button
+        if buy_button == None:
+            raise Error("Could not find {} button.".format(transaction))
+        buy_button.click()
+
+        if "Apology" in self.driver.title and outcome:
+            raise Error("Expected {} success, not failure.".format(transaction))
```
dmalan: "failure" seems vague for students, should probably more colloquially be "not an apology" or such?

checks/cs50/2017/x/finance/checks.py
```diff
@@ -1,13 +1,305 @@
...
+        self.server_up()
+        res = requests.get(Finance.BASE_PATH)
+        if res.status_code != 200:
+            raise Error((res.status_code, 200))
+        self.server_down()
+
+    @check("startup")
+    def can_register(self):
+        """user can register for account."""
+        self.server_up()
```
dmalan: Hm, should `server_up` and `server_down` really be the responsibility of each check? Should/could `startup` not do that? And maybe we could change `check` to take two arguments, the second of which has a default value of `None` but, if present, represents a function that should be called after this check is done, for cleanup?

checks/cs50/2017/x/finance/checks.py
```diff
@@ -1,13 +1,305 @@
...
+    def history(self):
+        """history shows transactions."""
+        self.server_up()
+        self.login("check50user1", "crimson")
+        self.transaction("buy", "BBBB", "1", True)
+        self.driver.get("{}/history".format(Finance.BASE_PATH))
+        if not re.search("14.00", self.driver.page_source):
+            raise Error("Expected to see result of purchase in history.")
+        if not re.search("28.00", self.driver.page_source):
+            raise Error("Expected to see result of sale in history.")
```
dmalan: Errors like this should probably be more explicit, a la "Expected to see result of sale (i.e., 28.00) ..."?

checks/cs50/2017/x/finance/lookup.py
```diff
@@ -0,0 +1,9 @@
+def lookup(symbol):
```
dmalan: Should comment files like this for readers' sake.

dmalan:
Really thrilled how well web testing integrates with `check50`, @brianyu28, really clean overall.

checks/cs50/2017/x/finance/checks.py
brianyu28: I had originally wanted to do this. Unfortunately, Selenium's API doesn't support accessing status codes:

https://github.com/seleniumhq/selenium-google-code-issue-archive/issues/141

dmalan: Why is that not a feature??

brianyu28: In my opinion, it really should be.

But in their words: "We will not be adding this feature to the WebDriver API as it falls outside of our current scope (emulating user actions)."

dmalan: 👎

checks/cs50/2017/x/finance/checks.py
crossroads1112: `self.replace_fn(asm_file, old_fn_name, new_fn_name)`

crossroads1112:
Superseded by #62

PR #20, opened 2017-06-29
Auto update check50 based on PyPi

### LINKED ISSUE #19: auto-update check50?
- run `pip install --upgrade` first
- then spawn a child process running `check50`, orphan it, kill the parent

brianyu28:
Resolves #19.

This checks pypi for the latest version of `check50` and compares it against the current version. If a newer version is available, `pip install --upgrade` is run to upgrade the version of `check50`, after which we re-run `check50` (this time passing in the `--no-autoupdate` flag).

The relevant additional code:

```python
    # check for newer version on PyPi
    pypi = pypijson.get("check50")
    if pypi and not args.no_autoupdate and StrictVersion(pypi["info"]["version"]) > VERSION:

        # updade check50
        pip = "pip3" if sys.version_info >= (3, 0) else "pip"
        subprocess.run([pip, "install", "--upgrade", "check50"])
        check50 = os.path.realpath(__file__)
        os.execv(check50, sys.argv + ["--no-autoupdate"])
        raise RuntimeError("Impossible.")
```

### DIFF
--- a/check50.py
+++ b/check50.py
@@ -43,8 +43,7 @@ def main():
 
     # check for newer version on PyPi
     pypi = pypijson.get("check50")
-    pypi_version = StrictVersion(pypi["info"]["version"])
-    if pypi and not args.no_autoupdate and pypi_version > VERSION:
+    if pypi and not args.no_autoupdate and StrictVersion(pypi["info"]["version"]) > VERSION:
 
         # updade check50
         pip = "pip3" if sys.version_info >= (3, 0) else "pip"
--- a/check50.py
+++ b/check50.py
@@ -34,7 +34,7 @@ def main():
     parser.add_argument("-d", "--debug", action="store_true")
     parser.add_argument("-l", "--local", action="store_true")
     parser.add_argument("--log", action="store_true")
-    parser.add_argument("--no-autoupdate", action="store_true")
+    parser.add_argument("--update", action="store_true")
     parser.add_argument("-v", "--verbose", action="store_true")
 
     args = parser.parse_args()
@@ -43,14 +43,14 @@ def main():
 
     # check for newer version on PyPi
     pypi = pypijson.get("check50")
-    if pypi and not args.no_autoupdate and StrictVersion(pypi["info"]["version"]) > VERSION:
+    if pypi and args.update and StrictVersion(pypi["info"]["version"]) > VERSION:
 
         # updade check50
         pip = "pip3" if sys.version_info >= (3, 0) else "pip"
         subprocess.run([pip, "install", "--upgrade", "check50"])
         check50 = os.path.realpath(__file__)
-        os.execv(check50, sys.argv + ["--no-autoupdate"])
-        raise RuntimeError("Impossible.")
+        sys.argv.remove("--update")
+        os.execv(check50, sys.argv)
 
     if not args.local:
         try:
--- a/check50.py
+++ b/check50.py
@@ -1,5 +1,7 @@
 #!/usr/bin/env python3
 
+from __future__ import print_function
+
 import argparse
 import hashlib
 import importlib
@@ -19,11 +21,10 @@ import unittest
 
 from distutils.version import StrictVersion
 from functools import wraps
+from pkg_resources import get_distribution, parse_version
 from termcolor import cprint
 
 import config
-
-VERSION = StrictVersion("2.0.0")
 
 def main():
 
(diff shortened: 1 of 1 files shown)


check50.py
```diff
@@ -32,8 +33,8 @@ def main():
     parser.add_argument("files", nargs="*")
     parser.add_argument("-d", "--debug", action="store_true")
     parser.add_argument("-l", "--local", action="store_true")
-    parser.add_argument("-f", "--force", action="store_true")
     parser.add_argument("--log", action="store_true")
+    parser.add_argument("--no-autoupdate", action="store_true")
```
dmalan: I wonder if we should flip this to be `--update` and then alias `check50` to `check50 --update` on CS50 IDE, omitting that alias from CS50 IDE Offline? In one or the other, we'll need to disable. Is either better?

check50.py
```diff
@@ -42,8 +43,14 @@ def main():
...
-    if pypi and not args.force and StrictVersion(pypi["info"]["version"]) > VERSION:
-        raise RuntimeError("You are running an old version of check50. Run pip install check50 --upgrade, and then run check50 again!")
+    if pypi and not args.no_autoupdate and StrictVersion(pypi["info"]["version"]) > VERSION:
+
+        # updade check50
+        pip = "pip3" if sys.version_info >= (3, 0) else "pip"
+        subprocess.run([pip, "install", "--upgrade", "check50"])
+        check50 = os.path.realpath(__file__)
+        os.execv(check50, sys.argv + ["--no-autoupdate"])
+        raise RuntimeError("Impossible.")
```
dmalan: What's the intent of this, to kill the parent? Why an exception and not `exit`?

brianyu28: `os.execv` replaces the current process (the old version of `check50`) with the new process (the new version of `check50`), which means that this line of code is unreachable: the new run of `check50` will replace the old one, and we start from the beginning again.

check50.py
brianyu28: I was just trying to account for the use case that a student just runs `check50` with no arguments and we automatically check for updates. But yeah, if we alias `check50` to `check50 --update`, that resolves the issue. Can flip this to `--update` if that works better for offline users (offline users, for what it's worth, will need `check50` aliased to `check50 --local`).

check50.py
dmalan: Is it conventional to have an arbitrary line of code like that, though? As opposed to nothing (or just a comment)?

check50.py
dmalan: Hm, am thinking `--update` is more natural, since it's more conventional to have to tell a program to update something, rather than it being assumed. And aliasing on IDE to `check50 --update` and on Offline to `check50 --local` feels like a nice parallel?

dmalan: Actually, I retract this. For folks who install `check50` on their own machines, it's indeed probably best if updating be the default behavior, else they'll likely never update. I do the same with `cli50` for that reason. For `cli50`, I used `-f`/`--fast`, which I borrowed from, I think, standard GNU arg conventions. (`-f` is easy to type too when I don't want to wait for the Docker image to update.) But since `check50` is server-side, not sure "fast" is the right connotation here. The `--no-something` convention is common. Or we could do `--update=X`, where `X` is documented as `1` by default? Not sure which is the more common convention?

dmalan:
0:-) https://github.com/cs50/check50/pull/20#discussion_r124712410

brianyu28:
Reverting back!

check50.py
brianyu28: I know I've seen impossible exceptions before, but didn't find much after looking around. Deleted the line for now as unnecessary.

check50.py
brianyu28: Reverted back to `--no-autoupdate`.

check50.py
```diff
@@ -42,8 +43,13 @@ def main():
 
     # check for newer version on PyPi
     pypi = pypijson.get("check50")
-    if pypi and not args.force and StrictVersion(pypi["info"]["version"]) > VERSION:
-        raise RuntimeError("You are running an old version of check50. Run pip install check50 --upgrade, and then run check50 again!")
+    if pypi and not args.no_autoupdate and StrictVersion(pypi["info"]["version"]) > VERSION:
```
dmalan: We should probably check if `check50` was actually installed as a package, as via https://github.com/cs50/submit50/blob/develop/submit50.py#L32, since we wouldn't want to try updating if a user has installed/run from source?

check50.py
```diff
@@ -42,8 +43,13 @@ def main():
...
+
+        # updade check50
+        pip = "pip3" if sys.version_info >= (3, 0) else "pip"
+        subprocess.run([pip, "install", "--upgrade", "check50"])
```
dmalan: If this fails, shouldn't `execv`, should probably just proceed without autoupdating and print warning to `stderr`?

check50.py
brianyu28: Updated!

check50.py
brianyu28: Updated!

PR #22, opened 2017-07-07
Pset4

hljames:


checks/cs50/2017/x/recover/checks.py
```diff
@@ -0,0 +1,68 @@
...
+    @check("compiles")
+    def test_noimage(self):
+        """handles lack of forensic image"""
+        self.spawn("./recover").exit(1)
+
+    @check("compiles")
+    def first_image(self):
+        """recovers 000.jpg correctly"""
+        self.include("card.raw")
+        subprocess.call("bash", shell=True)
```
dmalan: What's this for?

checks/cs50/2017/x/recover/checks.py
```diff
@@ -0,0 +1,68 @@
...
+        self.spawn("./recover card.raw").exit(0)
+        hash000 = 'c3ceda9dde5cf2212e96d6f61e2e849ac60e2c3cbdf0184f6e0c3b6ea5e69d31'
```
dmalan: Double quotes for consistency.

checks/cs50/2017/x/recover/checks.py
```diff
@@ -0,0 +1,68 @@
...
+        if self.hash("000.jpg") != hash000:
```
dmalan: Not sure we gain much by having the variable.

checks/cs50/2017/x/recover/checks.py
```diff
@@ -0,0 +1,68 @@
...
+            raise Error("recovered image does not match")
+
+    @check("compiles")
+    def middle_images(self):
+        """recovers middle images correctly"""
+        self.include("card.raw")
```
dmalan: @crossroads1112 should the semantics here perhaps be `self.copy` or `self.import` instead? Not sure, just had to think twice about what `include` meant here.

checks/cs50/2017/x/recover/checks.py
```diff
@@ -0,0 +1,68 @@
...
+        self.spawn("./recover card.raw").exit(0)
+        middlehashes = ['8c0bd07a6c232670fd55b50ac42611c349920a119ef575eb28f770b9c89753e4',
```
dmalan: Double quotes.

checks/cs50/2017/x/recover/checks.py
```diff
@@ -0,0 +1,68 @@
...
+            'f22fd0a4e18f076c043bd8306fe771f2fa843f764e5c9c399a8f61d4921c1743',
+            '5933fbd4d2ce7194ed67241c1f65fc86b5c882884de63644ceef4cd6ec97dd4c',
+            'cbf1f76713e3a0c6b2779f9a3489291783ee9e8f4ec7469d5be362e892cbfbdd',
+            '24b2b88477534ee29f7370d0742c172bef083a8fa78505cbc4bf7fd7df3353ca',
+            '2f1900342cfa54c0b5be9bc65d3dedbdcc2fe48c105b75700de1087ba93f581a',
+            'e13ddf43b9524f0cfdf23b8ac89b1bcdc21bb9030d5a80637a511f84c2c807e6',
+            'ff87370119a48b8f78b8ff96476bfb1f0a317339922eb6d526859674f1349bd7']
+        c = 0
+        while (c < 14):
+            h = self.hash("0" + str(c + 1).zfill(2) + ".jpg")
```
dmalan: This block should definitely be commented.

checks/cs50/2017/x/recover/checks.py
```diff
@@ -0,0 +1,68 @@
...
+            if (h != middlehashes[c]):
+                raise Error("recovered image does not match")
+            c += 1
+
+    @check("compiles")
+    def last_image(self):
+        """recovers 015.jpg correctly"""
+        self.include("card.raw")
+        self.spawn("./recover card.raw").exit(0)
+        hash015 = '49442513773931634fc572de401bb44d6d91069b23113e86a16336f07d69021d'
```
dmalan: Quotes.

checks/cs50/2017/x/recover/checks.py
```diff
@@ -0,0 +1,68 @@
...
+        if self.hash("015.jpg") != hash015:
```
dmalan: Var unneeded.

checks/cs50/2017/x/resize/less/checks.py
```diff
@@ -0,0 +1,65 @@
...
+
+sys.path.append(os.getcwd())
+from check50 import *
+
+class Resize_Less(TestCase):
+
+    @check()
+    def resize_bmp_exist(self):
+        """resize.c and bmp.h exist."""
+        super().exists("resize.c")
```
dmalan: @crossroads1112 can we change `exists` to take a `list` of files too?

checks/cs50/2017/x/resize/less/checks.py
```diff
@@ -0,0 +1,65 @@
...
+        child = self.spawn("./resize 1 1.bmp outfile.bmp")
+        self.spawn("diff 1.bmp outfile.bmp").stdout("", "").exit(0)
+
+    @check("compiles")
+    def scale_by_2(self):
+        """resizes 1x1-pixel BMP to 2x2 correctly when n is 2"""
+        self.include("1.bmp")
+        self.include("2.bmp")
+        child = self.spawn("./resize 2 1.bmp outfile.bmp")
+        self.spawn("diff 2.bmp outfile.bmp").stdout("", "").exit(0)
```
dmalan: Why expecting `""` here and elsewhere?

checks/fifteen.py
```diff
@@ -0,0 +1,82 @@
...
+        import pexpect
+        child = pexpect.spawn("./fifteen 3")
+        child.expect(".+")
+        import time
+        print(os.getcwd(), file=sys.stderr)
+        self.include("outputs/fifteen/init-3x3.txt")
+        # child  = subprocess.Popen("./fifteen 3", shell=True)
+        # time.sleep(3)
+        # child.kill()
+        subprocess.Popen("bash", shell=True).wait()
```
dmalan: What's this?

checks/cs50/2017/x/resize/less/checks.py
brianyu28: Already done per https://github.com/cs50/check50/commit/e2da3f2c0dcc4892f317d3c3f66749ba9696a188!

dmalan: Thanks, @crossroads1112! :)

checks/cs50/2017/x/recover/checks.py
crossroads1112: Copy feels semantically like it takes two arguments. `self.copy("card.raw")` seems to raise the question "Copy card.raw to/from where?"

Import doesn't feel quite right either. At least in the pythonic sense, `import` feels like it implies that I'm going execute `card.raw` somehow to facilitate the check. 

I think `include` is the right semantic. It's sort of saying that when gathering all the stuff we need for this check, we must `include card.raw`. It also evokes `#include` which is a sort of copy/paste operation.

checks/cs50/2017/x/resize/less/checks.py
crossroads1112: I believe it means we are expecting no output i.e. diff outputs nothing. The exit code is probably a better indicator though. `diff` returns a 0 when the files are identical, 1 when they're different or a 2 on error (file not found).

checks/fifteen.py
crossroads1112: IIRC, this is an artifact of debugging the check. Should be removed.

checks/cs50/2017/x/recover/checks.py
dmalan: 👍

checks/cs50/2017/x/resize/less/checks.py
dmalan: Didn't we introduce `.stdout(self.EOF)` or such for this?

checks/cs50/2017/x/recover/checks.py
crossroads1112: Artifact of debugging I think, will remove

PR #25, opened 2017-07-19
added setup.py

kzidane:
No need to deploy this to PyPI for now. Just to make it easier to install in the `Dockerfile` using `pip install .` or something.

### DIFF
--- a/setup.py
+++ b/setup.py
@@ -1,0 +1,23 @@
+from setuptools import setup
+
+setup(
+    author="CS50",
+    author_email="sysadmins@cs50.harvard.edu",
+    classifiers=[
+        "Intended Audience :: Education",
+        "Programming Language :: Python :: 3",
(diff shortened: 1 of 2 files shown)


setup.py
```diff
@@ -0,0 +1,23 @@
...
+    author_email="sysadmins@cs50.harvard.edu",
+    classifiers=[
+        "Intended Audience :: Education",
+        "Programming Language :: Python :: 3",
+        "Topic :: Education",
+        "Topic :: Utilities"
+    ],
+    description="This is check50, with which you can check solutions to \
+        problems for CS50.",
+    install_requires=[ "argparse", "pexpect", "requests", "termcolor" ],
```
dmalan: inconsistent spaces after `[` and before `]`.

kzidane: Oops, sorry! Removed per 95d3891!

PR #43, opened 2017-07-23
Pset5

hljames:


crossroads1112:
This PR removes many of the other checks, which is very much not what we want

checks/cs50/2017/x/speller/checks.py
```diff
@@ -0,0 +1,58 @@
...
+from check50 import *
+
+class Speller(Checks):
+
+    @check()
+    def exists(self):
+        """dictionary.c, dictionary.h, and Makefile exist"""
+        super(Speller, self).exists("dictionary.c")
+        super(Speller, self).exists("dictionary.h")
+        super(Speller, self).exists("Makefile")
```
dmalan: Is simpler syntax possible? Is this a Python 2/3 thing? If so, can we abstract this messiness away?

checks/cs50/2017/x/speller/checks.py
```diff
@@ -0,0 +1,58 @@
...
+
+    @check("exists")
+    def compiles(self):
+        """speller compiles"""
+        self.include("speller.c")
```
dmalan: So I'll admit each time I read `include` I have to think about the semantics, and I worry TFs might think this means to import its contents into the problem's own code (a la our append trick). Would `install` be even clearer?

checks/cs50/2017/x/speller/checks.py
```diff
@@ -0,0 +1,58 @@
...
+        self.spawn("make").exit(0)
+
+    @check("compiles")
+    def basic(self):
+        """handles most basic words properly"""
+        self.include("basic")
+        self.spawn("./speller basic/dict basic/text").stdout(File("basic/out")).exit(0)
```
dmalan: In the interests of simplicity, can you remind me what the issue was with calling this `self.run`?

checks/cs50/2017/x/speller/checks.py
```diff
@@ -0,0 +1,58 @@
...
+        self.include("max_length")
+        self.spawn("./speller max_length/dict max_length/text").stdout(File("max_length/out")).exit(0)
+
+    @check("compiles")
+    def apostrophe(self):
+        """handles words with apostrophes properly"""
+        self.include("apostrophe")
+        self.spawn("./speller apostrophe/without/dict apostrophe/with/text").stdout(File("apostrophe/outs/without-with")).exit(0)
+        self.spawn("./speller apostrophe/with/dict apostrophe/without/text").stdout(File("apostrophe/outs/with-without")).exit(0)
+        self.spawn("./speller apostrophe/with/dict apostrophe/with/text").stdout(File("apostrophe/outs/with-with")).exit(0)
```
dmalan: Does having this `out/` subdir make things clearer? Should those sample outputs not be inside he same `without/` and `with/` directories where the corresponding inputs are?

checks/cs50/2017/x/speller/checks.py
crossroads1112: This is a Python 2 thing. IMHO it would be much clearer to write `Checks.exists("dictionary.c")` (indeed  that's what stack overflow seems to suggest). I was planning on creating a PR once all the psets were done that made this change.

checks/cs50/2017/x/speller/checks.py
crossroads1112: Paging @brianyu28

checks/cs50/2017/x/speller/checks.py
crossroads1112: Perhaps the `out` dir is unnecessary. Will remove

checks/cs50/2017/x/speller/checks.py
dmalan: To avoid having contributors have to remember special syntax just for this one check, should we just rename the parent's method to `isfile`, a la https://docs.python.org/2/library/os.path.html#os.path.isfile? Then every check can just use `self.*`.

crossroads1112: I'm not sure if that solves the special syntax problem. Then we sort of have two methods which conceptually do pretty much the same thing with differing names. It's also a bit confusing since calling `isfile` on a directory is perfectly valid.

checks/cs50/2017/x/speller/checks.py
crossroads1112: IMHO of the alternatives, `import` is probably the clearest

checks/cs50/2017/x/speller/checks.py
brianyu28: Issue here is that we're subclassing `unittest.TestCase` which itself already has a method call `run` whose semantics is "run the test". That `run` method gets called when a `unittest.TestSuite` is running, so if we name our function `self.run`, we override the existing `run` method and `unittest` gets confused. So we needed something other than `run` as the name of the function, and `spawn` made the most sense to be consistent with `pexpect.spawn`.

checks/cs50/2017/x/speller/checks.py
crossroads1112: Actually, on second thought I think that it does. There are three outputs each of which are a different combination of using a with/without text and a with/without dict, so they don't really fit into either the `with` or `without` dir.

checks/cs50/2017/x/speller/checks.py
dmalan: Do we indeed need to descend from `unittest.TestCase`, or could an instance of that just be stored inside of `Check` as a "private" instance variable?

checks/cs50/2017/x/speller/checks.py
dmalan: Hm, though in Python, `import` is pretty similar in spirit, so far as TFs would perceive it, as `include` in C. I think we used to use `copy`? We could then support an optional second argument (destination file name) so that, if we have multiple `speller.c` files (in some other scenario), we can name them in some self-descriptive way but copy them into the student's dir as just `speller.c`?

checks/cs50/2017/x/speller/checks.py
brianyu28: We do need to be a descendant of `unittest.TestCase` otherwise `unittest` itself won't know which checks to run: just storing the instance as a private instance variable means that the instance wouldn't contain any checks to run, since checks are defined as methods of the `unittest.TestCase` subclass.

checks/cs50/2017/x/speller/checks.py
crossroads1112: Why could the check not just use the original self-descriptive name for its own purposes? It seems like we'd be adding a feature for the sake of semantics rather than utility. Would `require` work here?

dmalan: Hm, `require` could be assumed (by TFs) to mean "student must have a file called `speller.c` is all. I keep mentioning `copy` partly because that's how Docker does this; it's (essentially) a synonym for `add` in Docker too. How about that, `add`? Pretty unambiguous?

checks/cs50/2017/x/speller/checks.py
dmalan: k

checks/cs50/2017/x/speller/checks.py
crossroads1112: I like it, let's go with that. Optimally, I'd like for all or most of these PRs to be merged first so I can update all the checks in a single commit

dmalan: Aok, just assigned them all back to you after reviewing. https://github.com/cs50/check50/pulls

crossroads1112: Does that mean the `pset` PRs are okay to merge?

PR #46, opened 2017-07-24
Add pset0 Scratch checks

brianyu28:
By unzipping the Scratch `.sb2` file and parsing the resulting `.json` file, we're able to write test cases for the Scratch problem set as well now. In this test suite, I check for the project containing two sprites (at least one of which isn't the cat), three scripts, and at least one condition, loop, variable, and sound, which matches [our pset0 spec](http://docs.cs50.net/2017/x/psets/0/pset0.html) precisely save for the "Your project should be more complex than most of those demonstrated in lecture" requirement.

Successful output looks like:

```diff
Brian-MBP:check50 brian$ python check50.py --local 2017/x/scratch project.sb2
:) project exists and is valid Scratch program
:) project contains at least two sprites
:) project contains a non-cat sprite
:) project contains at least three scripts
:) project uses at least one condition
:) project uses at least one loop
:) project uses at least one variable
:) project uses at least one sound
```

### DIFF
--- a/checks/cs50/2017/x/scratch.py
+++ b/checks/cs50/2017/x/scratch.py
@@ -1,0 +1,130 @@
+import json
+from check50 import *
+
+
+class Scratch(Checks):
+
+    @check()
+    def valid(self):
+        """project exists and is valid Scratch program"""
+        self.exists("project.sb2")
+
+        # Ensure that unzipped .sb2 file contains .json file
+        self.spawn("unzip project.sb2").exit()
+        self.exists("project.json")
+
+    @check("valid")
+    def two_sprites(self):
+        """project contains at least two sprites"""
+        project = json.loads(File("project.json").read())
+
+        # Loop over children: includes sprites and stages.
+        num_sprites = sum("costumes" in child for child in project["children"])
+
+        if num_sprites < 2:
+            raise Error("Only {} sprite{} found, 2 required.".format(num_sprites,
+                "" if num_sprites == 1 else "s"))
+
+    @check("valid")
+    def non_cat(self):
+        """project contains a non-cat sprite"""
+        project = json.loads(File("project.json").read())
+        non_cat = False
+
+        # As a heuristic, only the cat has a "meow" sound.
+        for child in project["children"]:
+
+            # Skip over any non-sprites (e.g. backdrops).
+            if "costumes" not in child:
+                continue
+
+            # Check if the sprite has a "meow" sound.
+            can_meow = any(sound["soundName"] == "meow" for sound in child.get("sounds", []))
+
+            # If it doesn't meow, we've found a non-cat sprite.
+            if not can_meow:
+                return
+
+        # If we haven't returned, then no non-cat sprite found.
+        raise Error("Requires a non-cat sprite.")
+
+    @check("valid")
+    def three_scripts(self):
+        """project contains at least three scripts"""
+        project = json.loads(File("project.json").read())
+
+        # Add up scripts from each sprite or backdrop.
+        num_scripts = sum(len(child.get("scripts", [])) for child in project["children"])
+
+        if num_scripts < 3:
+            raise Error("Only {} script{} found, 3 required.".format(num_scripts,
+                "" if num_scripts == 1 else "s"))
+
+    @check("valid")
+    def uses_condition(self):
+        """project uses at least one condition"""
+        project = json.loads(File("project.json").read())
+
+        # Search project scripts for an if or if/else block.
+        if not project_contains_keywords(project, ["doIf", "doIfElse"]):
+            raise Error("No conditions found, 1 required.")
+
+    @check("valid")
+    def uses_loop(self):
+        """project uses at least one loop"""
+        project = json.loads(File("project.json").read())
+
+        # Search project scripts for a repeat, repeat until, or forever block.
+        if not project_contains_keywords(project, ["doRepeat", "doUntil", "doForever"]):
+            raise Error("No loops found, 1 required.")
+
+    @check("valid")
+    def uses_variable(self):
+        """project uses at least one variable"""
+        project = json.loads(File("project.json").read())
+
+        # Look for global variables.
+        if project.get("variables"):
+            return
+
+        # Look for local-to-sprite variables.
+        if not any(child.get("variables") for child in project["children"]):
+            return
+
+        # If we've reached this point, no variable found.
+        raise Error("No variables found, 1 required.")
+
+    @check("valid")
+    def uses_sound(self):
+        """project uses at least one sound"""
+        project = json.loads(File("project.json").read())
+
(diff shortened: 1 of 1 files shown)


checks/cs50/2017/x/scratch/checks.py
```diff
@@ -0,0 +1,155 @@
...
+        self.exists("project.sb2")
+
+        # Ensure that unzipped .sb2 file contains .json file
+        self.spawn("unzip project.sb2").exit()
+        self.exists("project.json")
+    
+    @check("valid")
+    def two_sprites(self):
+        """project contains at least two sprites"""
+        project = json.loads(open("project.json", "r").read())
```
crossroads1112: `File("project.json").read()` is probably better here

checks/cs50/2017/x/scratch/checks.py
```diff
@@ -0,0 +1,155 @@
...
+                continue
+            
+            # If sprite has no sounds, it's probably not a cat.
+            can_meow = False
+            if "sound" not in child:
+                return
+
+            for sound in child["sounds"]:
+                if sound["soundName"] == "meow":
+                    can_meow = True
```
crossroads1112: `can_meow = any(sound["soundName"] == "meow" for sound in child["sounds"])` is a sexier/slightly more efficient way to do this

checks/cs50/2017/x/scratch/checks.py
```diff
@@ -0,0 +1,155 @@
...
+    def three_scripts(self):
+        """project contains at least three scripts"""
+        project = json.loads(File("project.json").read())
+        
+        num_scripts = 0
+
+        # Add up scripts from each sprite or backdrop.
+        for child in project["children"]:
+            if "scripts" in child:
+                num_scripts += len(child["scripts"])
```
crossroads1112: In a similar vein, `num_scripts = sum(len(child.get("scripts", [])) for child in project["children"])` is a little sexier

checks/cs50/2017/x/scratch/checks.py
```diff
@@ -0,0 +1,155 @@
...
+        """project uses at least one variable"""
+        project = json.loads(File("project.json").read())
+
+        # Look for global variables.
+        if "variables" in project and len(project["variables"]) > 0:
+            return
+        
+        # Look for local-to-sprite variables.
+        for child in project["children"]:
+            if "variables" in child and len(child["variables"]) > 0:
```
crossroads1112: I think its a bit more pythonic to do `len(child.get("variables", [])) > 0`

checks/cs50/2017/x/scratch/checks.py
```diff
@@ -0,0 +1,155 @@
...
+
+    # Iterate over all sprites and backdrops.
+    for child in project["children"]:
+        if "scripts" not in child:
+            continue
+        
+        # Perform a DFS on each script looking for keywords.
+        for script in child["scripts"]:
+            if contains(script, keywords):
+                return True
```
crossroads1112: Similarly, could probably do `any(contains(script, keywords) for script in child.get("scripts", []))`

checks/cs50/2017/x/scratch/checks.py
```diff
@@ -0,0 +1,155 @@
...
+    if type(script) != list:
+        return False
+    
+    if script[0] in keywords:
+        return True
+
+    # Iterate over all children.
+    for child in script:
+        if contains(child, keywords):
+            return True
```
crossroads1112: `return any(contains(child, keywords) for child in script)`

checks/cs50/2017/x/scratch/checks.py
```diff
@@ -0,0 +1,155 @@
...
+        """project contains at least two sprites"""
+        project = json.loads(File("project.json").read())
+        num_sprites = 0
+
+        # Loop over children: includes sprites and stages.
+        for child in project["children"]:
+            # Only sprites have costumes.
+            if "costumes" in child:
+                num_sprites += 1
+
```
crossroads1112: `sum("costumes" in child for child in project["children"])`

checks/cs50/2017/x/scratch/checks.py
```diff
@@ -0,0 +1,139 @@
...
+        # As a heuristic, only the cat has a "meow" sound.
+        for child in project["children"]:
+        
+            # Skip over any non-sprites (e.g. backdrops).
+            if "costumes" not in child:
+                continue
+            
+            # If sprite has no sounds, it's probably not a cat.
+            if "sounds" not in child:
+                return
```
crossroads1112: `get` takes care of this below, not necessary

checks/cs50/2017/x/scratch/checks.py
```diff
@@ -0,0 +1,139 @@
...
+        if any(contains(script, keywords) for script in child.get("scripts", [])):
+            return True
+
+    return False
+
+def contains(script, keywords):
+    """Performs DFS on the script to determine if keyword exists."""
+
+    # The keyword must be the first item in a list.
+    if type(script) != list or len(script) == 0:
```
crossroads1112: I think `not script` is more idiomatic than `len(script) == 0`

checks/cs50/2017/x/scratch/checks.py
```diff
@@ -0,0 +1,139 @@
...
+            if "sounds" not in child:
+                return
+
+            can_meow = any(sound["soundName"] == "meow" for sound in child.get("sounds", []))
+
+            # If it doesn't meow, we've found a non-cat sprite.
+            if not can_meow:
+                return
+        
+        # If we haven't returned, then no non-cat sprite found.
```
crossroads1112: This might be a reasonable place to use a `for/else` (whatever you think is clearer)

i.e.

```python
for child in project["children"]:

   # Skip over any non-sprites (e.g. backdrops).
    if "costumes" not in child:
        continue

   # If it doesn't meow, we've found a non-cat sprite
    if not any(sound["soundName"] == "meow" for sound in child.get("sounds, [])):
        break
else:
    raise Error("Requires a non-cat sprite")

**Comment by brianyu28** [comment]
Sample outputs:

screen shot 2017-07-25 at 8 17 16 am

**Comment by dmalan** [comment]
This is a great new feature for `check50`. @dlloyd09

File: checks/cs50/2017/x/scratch/checks.py [hunk 11]
```diff
@@ -0,0 +1,133 @@
+import json
+import os
+import sys
+
+sys.path.append(os.getcwd())
+from check50 import *
+
+class Scratch(Checks):
```
dmalan: @crossroads1112, we recently discussed adding a `@checks` decorator so that we needn't just assume the first (or only) class in a file is the one we want for checks. But now that we're adding `__init__.py`, is that moot, since we can rely on `__init__.py` to export the class we care about?

PR #47, opened 2017-07-25
Simplified mario checks

crossroads1112:
Wanted to get input on this. It seemed wrong to have two identical `checks.py` files for more/less comfortable mario especially when they had the same parent package. This PR takes advantage of the `__init__.py` that they share to greatly reduce the amount of redundancy.


checks/cs50/2017/x/mario/mario.py
```diff
@@ -0,0 +1,76 @@
...
+import sys
+
+sys.path.append(os.getcwd())
+from check50 import *
+
+class Mario(Checks):
+    @check()
+    def exists(self):
+        """mario.c exists."""
+        super(Mario, self).exists("mario.c")
```
dmalan: Did we decide this is the optimal Python 2/3 approach? I worry this is an API detail that a TF will always have to look up.

checks/cs50/2017/x/mario/mario.py
```diff
@@ -0,0 +1,76 @@
...
+
+    @check("exists")
+    def compiles(self):
+        """mario.c compiles."""
+        self.spawn("clang -o mario mario.c -lcs50").exit(0)
+
+    @check("compiles")
+    def test_reject_negative(self):
+        """rejects a height of -1"""
+        self.spawn("./mario").stdin("-1").reject().kill()
```
dmalan: Is `kill` necessary, or could our post-check teardown just kill and still-open `Child`?

checks/cs50/2017/x/mario/mario.py
```diff
@@ -0,0 +1,76 @@
...
+def check_pyramid(output, correct):
+    if output == correct:
+        return
+
+    err = Error((output, correct))
+    output = output.split("\n")
+    correct = correct.split("\n")
+
+    # check for trailing whitespace
+    if len(output) == len(correct) and all(ol.rstrip() == cl for ol, cl in zip(output, correct)):
```
dmalan: Let's definitely explain complexity like this in more of a comment for future TFs.

checks/cs50/2017/x/mario/more/checks.py
```diff
@@ -1,79 +1,5 @@
...
-    err = Error((output, correct))
-    output = output.split("\n")
-    correct = correct.split("\n")
-
-    # check for trailing whitespace
-    if len(output) == len(correct) and all(ol.rstrip() == cl for ol, cl in zip(output, correct)):
-        err.helpers = "Did you add too much trailing whitespace to the end of your pyramid?"
-    raise err
+class MarioMore(Mario):
+    pass
```
dmalan: Should this now be in a `mario/more/more.py` file?

checks/cs50/2017/x/mario/mario.py
crossroads1112: We didn't decide on what the best route was so I did it this way to resemble the other checks. Personally, I think `Checks.exists` is just as clear as `super().exists` when we have single inheritance (that's what StackOverflow seems to suggest anyway).

checks/cs50/2017/x/mario/more/checks.py
crossroads1112: Per our discussion in slack, this approach enables certain slugs which are not intended. In any case, I wanted to do big changes like this all at once in a single commit after the PRs were merged so there aren't a bunch of unnecessary commits laying around.

checks/cs50/2017/x/mario/mario.py
crossroads1112: 566590c

PR #51, opened 2017-07-31
Deploy check50 v2.0.0

brianyu28:


setup.py
```diff
@@ -0,0 +1,24 @@
...
+        "Programming Language :: Python :: 3",
+        "Topic :: Education",
+        "Topic :: Utilities"
+    ],
+    description="This is check50, with which you can check solutions to \
+        problems for CS50.",
+    install_requires=["argparse", "pexpect", "requests", "backports.shutil_which", "termcolor"],
+    keywords=["check", "check50"],
+    name="check50",
+    py_modules=["check50", "config"],
```
dmalan: Do we (still) need `config`?

setup.py
```diff
@@ -0,0 +1,24 @@
...
+    author_email="sysadmins@cs50.harvard.edu",
+    classifiers=[
+        "Intended Audience :: Education",
+        "Programming Language :: Python :: 3",
+        "Topic :: Education",
+        "Topic :: Utilities"
+    ],
+    description="This is check50, with which you can check solutions to \
+        problems for CS50.",
+    install_requires=["argparse", "pexpect", "requests", "backports.shutil_which", "termcolor"],
```
dmalan: Requires `submit50` no?

brianyu28: Added!

setup.py
brianyu28: Yeah, per https://docs.python.org/3/faq/programming.html#how-do-i-share-global-variables-across-modules, it's still the canonical way to share variables between different modules.

dmalan: k

PR #52, opened 2017-07-31
Shorten output URL

brianyu28:


check50.py
```diff
@@ -94,7 +94,7 @@ def main():
                 # Terminate if no response.
                 if pings > 45:
                     cprint("check50 is taking longer than normal!", "red", file=sys.stderr)
-                    cprint("more info at: https://cs50.me/check50/results/{}/{}".format(username, commit_hash), "red", file=sys.stderr)
+                    cprint("more info at: https://cs50.me/checks/{}".format(commit_hash), "red", file=sys.stderr)
```
dmalan: For consistency with v1, can we just do `See https://cs50.me/checks/{} for more detail.`?

check50.py
```diff
@@ -112,7 +112,7 @@ def main():
 
             # Print results from payload.
             print_results(payload["checks"], config.args.log)
-            print("detailed results: https://cs50.me/check50/results/{}/{}".format(username, commit_hash))
+            print("detailed results: https://cs50.me/checks/{}".format(commit_hash))
```
dmalan: Here too?

PR #53, opened 2017-08-01
Skip check on check50 error

brianyu28:
So right now, when check50 runs into an error (which has happened a couple times now, usually because of students' code throwing an exception that we don't catch), `unittest` calls the `addError` function.

Currently, `addError` prints out "check50 ran into an error while running checks!" and just terminates the program. But that's not actually all that helpful to the students-- because the checks never finish running and they never get to see the results. So instead, this marks the check for which check50 hit an error as "skipped", and appends to the log details about the exception that was thrown.

### DIFF
--- a/check50.py
+++ b/check50.py
@@ -338,18 +338,17 @@ class TestResult(unittest.TestResult):
         })
 
     def addError(self, test, err):
+        test.log.append(err[1])
+        test.log += traceback.format_tb(err[2])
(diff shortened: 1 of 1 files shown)


brianyu28:
Yeah, in this model, exiting with 0 is "passing" the check, exiting with non-0 is "failing" (assuming that we have a `.exit(0)` call in the check), and causing an exception in `check50.py` is "skipping" the check because there was some problem.

PR #62, opened 2017-08-27
Add Flask testing API

crossroads1112:
See https://github.com/cs50/checks/pull/4 for checks written in the API.

We should also consider adding a way to denote that a check requires a certain version of `check50` (to prevent confusing errors if people try to run an old version of `check50` on their finance project.

### DIFF
--- a/check50.py
+++ b/check50.py
@@ -10,7 +10,6 @@ import inspect
 import json
 import os
 import pexpect
-import pip
 import re
 import requests
 import shutil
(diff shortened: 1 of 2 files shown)


crossroads1112:
@dmalan Is this good to go?

dmalan:
@brianyu28 want to try this with the other checks too, just for good measure?

brianyu28:
I've tried this with the finance checks, seems to work!

dmalan:
thanks!

PR #70, opened 2017-09-20
Adds segmentation fault detection, other bugfixes

crossroads1112:
It was reported that `check50` was incorrectly passing a student's find implementation since their program segfaulted when trying to find an element that didn't exist. Programs return 1 when they segfault, so it looked like the correct behavior. This PR alters the `Child.wait` method to check if the child was signaled with a `SIGSEGV` and raises an `Error` if it was.

It also fixes the bug where `checksdir` was being saved as a list and it wasn't expanding relative paths e.g. ".."

### DIFF
--- a/check50.py
+++ b/check50.py
@@ -14,6 +14,7 @@ import pip
 import re
 import requests
 import shutil
+import signal
 import subprocess
 import sys
 import tempfile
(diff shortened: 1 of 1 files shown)


check50.py
```diff
@@ -553,6 +552,10 @@ def wait(self, timeout=1):
 
         self.output = "".join(self.output).replace("\r\n", "\n").lstrip("\n")
         self.kill()
+
+        if self.child.signalstatus == signal.SIGSEGV:
+            raise Error("failed to execute program due to segmentation fault")
```
dmalan: Hm, technically executed, it just crashed? Perhaps more simply "segmentation fault while running program"? Though do we have access to program name in this context to be more explicit?

PR #74, opened 2017-09-29
adds .data, fixes --checkdir

dmalan:


setup.py
```diff
@@ -60,5 +60,5 @@ class CustomInstall(install):
         "console_scripts": ["check50=check50:main"]
     },
     url="https://github.com/cs50/check50",
-    version="2.0.6"
+    version="2.0.8"
```
kzidane: Why `8`?

dmalan: Actually, let's wait before merging, couple other changes pending it seems. And we should go to 2.1.0 anyway.

kzidane: :+1:

dmalan:
Okay, ready for you, @kzidane!

check50.py
```diff
@@ -67,6 +67,8 @@ def main():
                         help="display the full tracebacks of any errors")
 
     config.args = parser.parse_args()
+    if isinstance(config.args.checkdir, list):
```
crossroads1112: Since the `nargs=1` line was removed, how could this ever be a list?

crossroads1112:
This needs to be merged into `submit50` first: https://github.com/cs50/submit50/pull/147

PR #75, opened 2017-09-29
Added SIGINT handler

crossroads1112:
Fixes the issue where hitting `Ctrl-C` at the Github prompt, caused an "email sysadmins" message.

### DIFF
--- a/check50.py
+++ b/check50.py
@@ -84,6 +84,9 @@ def main():
             raise InternalError(
                 "submit50 is not installed. Install submit50 and run check50 again.")
         else:
+            submit50.handler.type = "check"
+            signal.signal(signal.SIGINT, submit50.handler)
+
             submit50.run.verbose = config.args.verbose
             username, commit_hash = submit50.submit("check50", identifier)
 
@@ -202,7 +205,7 @@ sys.excepthook = excepthook
 
 
 def handler(number, frame):
-    print()
+    termcolor.cprint("Check cancelled.", "red")
     sys.exit(1)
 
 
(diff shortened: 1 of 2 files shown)


check50.py
```diff
@@ -200,6 +201,11 @@ def excepthook(cls, exc, tb):
 sys.excepthook = excepthook
 
 
+def handler(number, frame):
+    print()
+    sys.exit(1)
+
+
```
dmalan: This seems to omit much of the handling we baked into `submit50`? Could we instead invoke the existing `handler` in `submit50.py` by exporting it (and parameterizing its `cprint` call to be "Submission cancelled." or "Checking cancelled."?

```
def handler(number, frame):
    """Handle SIGINT."""
    os.system("stty sane 2> {}".format(os.devnull))  # in case signalled from input_with_prefill
    if progress.progressing:
        progress(False)
    else:
        cprint()
    try:
        clear_credentials()
    except Exception:
        pass
    teardown()
    cprint(_("Submission cancelled."), "red")
    os._exit(0)
```

crossroads1112: We definitely can, although `submit50` isn't guaranteed to be imported since `check50` might be running locally. In which case, what would you want to print?

dmalan: Very true. Should we tweak `submit50.py` to change its output message based on `ORG`, as we do elsewhere? And then have `check50.py` call `submit50.py`'s `handler` if `config.args.local` is unset, else it uses its own and just prints "Checking cancelled." too? Slightly redundant but at least thorough for now.

PR #80, opened 2017-10-04
Expand check path

crossroads1112:


brianyu28:
@crossroads1112, I don't think this fixes the issue? I'm still experiencing it, at least.

Should it not be the other way around? e.g.

```
os.path.abspath(os.path.expanduser(config.args.checkdir))
```

instead of 

```
os.path.expanduser(os.path.abspath(config.args.checkdir))
```

else `~` is expanded to `path/to/cwd/~` first by `abspath`?

Issue #85, opened 2017-10-22
implementation of reject() not ideal for Python I/O


dmalan:
Because `.reject` sends `""`, the below fails with a `ValueError`, https://cs50.me/checks/67d9247dd25379fc1c646168a2eff04f10658314.

Sample input for `cs50/2017/fall/sentimental/mario/less`, per @curiouskiwi:

```
def main():

    while True:
       #height = get_int("Height: ")
       height = int(input("Height: "))
       if height < 24 and height >= 0:
           break

if __name__ == "__main__":
    main()
```

dmalan:
@crossroads1112, how best to redress, do you think?

CC @brianyu28 @dlloyd09 @curiouskiwi

crossroads1112:
I can't recall exactly why we implemented `reject` that way, but it seems like it could be accomplished in a less hackish way by just using `self.wait` with a short timeout. Something like:

```python
def wait(timeout=1):
    try:
        self.child.expect(".+", timeout=timeout)
    except TIMEOUT:
        raise Error("timed out while waiting for input prompt")
    except EOF:
        raise Error("expected prompt for input, found none")
    
    try:
        self.wait(timeout=1)
    except TIMEOUT:
        pass
    else:
        self.test.fail()
    
    return self
```

Thoughts, @brianyu28?

brianyu28:
Yeah that seems reasonable, though waiting even without the expectation of `.+` is probably sufficient too?

The price we pay for this approach, though, is that checks which test for rejection take longer, right? e.g. Mario, which tests for rejection four times, would have a check that takes four seconds longer (as opposed to it taking ~2 seconds right now locally). And I’m not sure I’d feel comfortable using less than a 1 second timeout because we’ve seen cases where due to ECS load, we sometimes hit short timeouts even when under normal conditions we wouldn’t.

If we’re willing to accept the additional time to check (or willing to take the risk of cutting down on the timeout to something shorter), then I think this makes sense? Once we parallelize checks, that might make the time issue here less important.

brianyu28:
Ah, didn't see @crossroads1112's edit the first time around, that's consistent with the first part of my above comment.

crossroads1112:
True, this makes check50 run significantly slower, though it does not require a prompt.


It appears this speed of the current approach may come at the cost of correctness though. For example, the problem passes the "rejects -1" case even though it clearly doesn't:

```c
#include <stdio.h>
#include <cs50.h>

int main(void)
{

    int n;
    do {
[...]

PR #86, opened 2017-10-23
Reimplement reject

crossroads1112:
Reimplements `Child.reject` as described in #85.

## Pros:
* Potentially more correct than the current approach (as demonstrated in the discussion in #85).
* Resolves the Python exception issues detailed in #85 

## Cons:
* Optimizes the case when the student program incorrectly accepts an input. Correct student code will result in waiting for the full timeout (default 1 second) to expire. This has potential negative performance implications if the majority of student code handles these cases correctly (which seems likely).


dmalan:
@brianyu28 mind reviewing/merging if all looks good?

brianyu28:
So long as we're willing to accept that checks might take a little longer, this looks good to me

PR #90, opened 2017-11-06
Temporary fix for flask exception handling.

crossroads1112:
cs50/checks#34

### DIFF
--- a/check50.py
+++ b/check50.py
@@ -527,8 +527,8 @@ class App(object):
             self.response = getattr(self.client, method.lower())(route, data=data, **kwargs)
         except BaseException as e:  # Catch all exceptions thrown by app
             # TODO: Change Finance starter code for edX and remove this as well as app.testing = True in __init__
-            self.test.log.append("exception raised in application: {}".format(e))
-            raise Error("application raised an exception (see log for more details)")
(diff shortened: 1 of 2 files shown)


crossroads1112:
The correct solution to this problem requires changing the starter code (per the linked issue) but its too late to do this for fall. This fixes the issue in the meantime.

dmalan:
@brianyu28 if we merge this, do we need to update cs50/ecs, or, @kzidane, will that happen automatically? Mind merging when ready?

brianyu28:
We'll need to first merge `develop` into `master` to trigger an update on PyPi, then we'll need to re-pull the Docker image from ECS.

brianyu28:
@kzidane, sound right to you? And if we do this now, do we run the risk of disrupting anyone's checks/submissions if ECS tasks are triggered during the pull?

crossroads1112:
Just realized, we should catch `BaseException` so that e.g. `SystemExit` thrown in the app doesn't exit check50

kzidane:
@dmalan won't happen automatically but I'll write a script to do that so when we deploy a new version of *50, it would trigger a build of our base image using Travis API.

@brianyu28 correct. Though based on previous experiences :innocent: I think we should wait 'til after the deadline?

dmalan:
Yes, after noon please!

PR #93, opened 2018-02-15
Lets remove some boilerplating! class X(Checks) in checks

Jelleas:
We can bind the methods decorated with `@check` dynamically to an instance of `Checks`. This way you don't have to nest `@check` methods in some class that inherits from `Checks`.

See: https://github.com/Jelleas/checks/blob/master/cs50/2017/x/hello/check50/__init__.py
Try: `python3 check50.py cs50/2017/x/hello@jelleas/checks --local`

### DIFF
--- a/check50.py
+++ b/check50.py
@@ -2,6 +2,7 @@
 
 from __future__ import print_function
 
+import types
 import argparse
 import errno
 import hashlib
@@ -135,10 +136,12 @@ def main():
 
     checks = import_checks(identifier)
 
-    # Create and run the test suite.
+    # Create the test suite
     suite = unittest.TestSuite()
     for case in config.test_cases:
         suite.addTest(checks(case))
+
+    # Run the test suite
     result = TestResult()
     suite.run(result)
     cleanup()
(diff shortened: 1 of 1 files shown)


dmalan:
Thanks!

check50.py
```diff
@@ -313,10 +315,18 @@ def import_checks(identifier):
     try:
         # Import module from file path directly.
         module = imp.load_source(slug, os.path.join(config.check_dir, "__init__.py"))
-        # Ensure that there is exactly one class decending from Checks defined in this package.
-        checks, = (cls for _, cls in inspect.getmembers(module, inspect.isclass)
-                   if hasattr(cls, "_Checks__sentinel")
-                   and cls.__module__.startswith(slug))
+        test_methods = [method for method in module.__dict__.values() if getattr(method, "is_check", False)]
```
crossroads1112: No reason to waste memory with a list comprehension here. Could use a generator comprehension but it's probably just simpler to replace this and the loop below with e.g.,

    for check in module.__dict__.values():
        if getattr(check, "is_check", False):
            setattr(CustomChecks, check.__name__, check)

check50.py
```diff
@@ -386,14 +396,16 @@ def wrapper(self):
...
-
 # Decorator for checks
 def check(dependency=None):
     def decorator(func):
 
         # add test to list of test, in order of declaration
         config.test_cases.append(func.__name__)
 
+        # set sentinel signalling this is a check
+        func.is_check = True
```
crossroads1112: I'd probably use `_is_check` here instead since this is a private method.

crossroads1112:
Closing the new version of check50 addresses this.

PR #108, opened 2018-07-18
update to version 3

cmlsharp:
Propose merging this into `develop` (though not `master` until January). This way the documentation at https://cs50.github.io/check50 matches the default branch. Any hotfixes we need to push to 2.2.3 can be done by branching off master.


crossroads1112:
@dmalan Per our discussion on friday, could we please merge this and cs50/submit50#166

PR #114, opened 2018-08-01
triggering manual50 build after deploy

kzidane:


.travis.yml
```diff
@@ -33,6 +33,14 @@ deploy:
...
   on:
     branch: master
+after_deploy: |
+  curl --silent --request POST \
+      --header "Content-Type: application/json" \
+      --header "Accept: application/json" \
+      --header "Travis-API-Version: 3" \
+      --header "Authorization: token $TRAVIS_TOKEN" \
+      --data '{"request": {"branch": "master"}}' \
+      https://api.travis-ci.com/repo/cs50%2Fmanual50/requests
```
dmalan: Is there no simpler way to trigger a build? Feels like this should be a simple key/value they support.

kzidane: Unfortunately there aren't any official wrappers in other languages. Just the REST API. I've emailed them and asked whether they plan to add support for build dependencies but haven't heard back.

I just realized we don't deploy this branch so I'll probably move to a `deploy` step and a script a la cs50/manual50 for now.

PR #133, opened 2018-08-29
add sanity checks to register

crossroads1112:
It occurred to me when writing the `Challenges` section, that weird things could happen if you called `after_check` when no check is being run or `{before,after}_every` when one is. This PR adds sanity checks to prevent that.

### DIFF
--- a/check50/__main__.py
+++ b/check50/__main__.py
@@ -29,10 +29,6 @@ from .runner import CheckRunner, CheckResult
 lib50.api.LOCAL_PATH = "~/.local/share/check50"
 
 
-class Error(Exception):
-    pass
-
-
 def excepthook(cls, exc, tb):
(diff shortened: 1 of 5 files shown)


Jelleas:
I like it! Small change though, instead of setting `internal.check_running` in `runner.py`, lets use our internal mechanism (`internal.register`) to set it.

crossroads1112:
@Jelleas It was intentional that this was outside the register mechanism. For example, there is currently a bit of an inconsistency insofar as if in an `after_every` callback, you could register another `after_every` callback (which is bad) since the `_not_running_callback` gets run first.

crossroads1112:
@Jelleas

crossroads1112:
merged this into devdev

PR #136, opened 2018-09-23
use lib50.config.Loader instead of lib50.config.load

Jelleas:
Prepares check50 for: <https://github.com/cs50/lib50/pull/11>

### DIFF
--- a/check50/__main__.py
+++ b/check50/__main__.py
@@ -26,7 +26,7 @@ from . import internal, __version__, simple, api
 from .api import Failure
 from .runner import CheckRunner, CheckResult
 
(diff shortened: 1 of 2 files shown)


kzidane:
@Jelleas do you wanna look at the conflicts here before reviewing and merging?

Jelleas:
This PR is the consequence of a proposed [big change in `lib50`](https://github.com/cs50/lib50/pull/11), moving to a configurable `files` option in `.cs50.yml`. However, the code in `lib50` is somewhat complex and the whole idea could use a second pair of eyes. I know @crossroads1112 has an opinion on this.

That said because it is a big and breaking change that affects submit50/check50/lab50, it might be wise to leave this open for summer.

PR #151, opened 2019-03-21
use subprocess to call pip instead of the removed pip.main() interface

stgm:
**tl;dr** - fix for failing finance checks. My students will be using next week, so if at all possible, I would appreciate a deploy 😬

It seems that in the IDE at ide.legacy.cs50.io, the checks for Finance aren't working anymore. These checks depend on an install via a `requirements.txt`. Apparently the recent version of `pip` that is now available in the IDE is the root cause of this failure.

`check50` currently uses the `main()` method from the pip module. However, this has been removed a couple of versions ago, [according to this](https://stackoverflow.com/questions/12332975/installing-python-module-within-code/50255019#50255019).

In this PR, I've replaced that call with the use of the `subprocess` module, which was already imported anyway. I have tested this change in the IDE, with the Finance checks, and they once again work. I have not tested this change with any other checks.

Obv this needs to be deployed to the IDE to be of any use :-)

### DIFF
--- a/check50.py
+++ b/check50.py
@@ -309,7 +309,7 @@ def import_checks(identifier):
                 args += ["--quiet"] * 3
 
             try:
-                code = __import__("pip").main(args)
+                code = subprocess.call([sys.executable, "-m", "pip", *args])
             except SystemExit as e:
                 code = e.code
 
--- a/check50.py
+++ b/check50.py
@@ -309,11 +309,8 @@ def import_checks(identifier):
                 args += ["--quiet"] * 3
 
             try:
-                code = subprocess.call([sys.executable, "-m", "pip", *args])
-            except SystemExit as e:
-                code = e.code
-
-            if code:
+                subprocess.check_call([sys.executable, "-m", "pip"] + args)
+            except subprocess.CalledProcessError:
                 raise InternalError("failed to install dependencies in ({})".format(
                     requirements[len(config.args.checkdir) + 1:]))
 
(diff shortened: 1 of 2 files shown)


stgm:
I see that the travis build on Python 3.4 is failing. I have used the 3.6 that is on the IDE to test this.

Apparently my * syntax is 3.5+ only, I'll let you decide if this is acceptable.

crossroads1112:
Yep, this is fixed in v3 but not in v2 at present. Currently the way you've done it is going to leave some dead code, so I'm going to recommend replacing the following:

```
            try:
                code = subprocess.call([sys.executable, "-m", "pip", *args])
            except SystemExit as e:
                code = e.code

            if code:
                raise InternalError("failed to install dependencies in ({})".format(
                    requirements[len(config.args.checkdir) + 1:]))
```

with 

```
            try:
                code = subprocess.check_call([sys.executable, "-m", "pip", *args])
            except subprocess.CalledProcessError:
                raise InternalError("failed to install dependencies in ({})".format(
                    requirements[len(config.args.checkdir) + 1:]))
```

With that fixed, I'll approve.

crossroads1112:
Regarding the failing checks, `check50` version 2 does need to work on python >= 2.7. v3 is 3.6 only, but until we roll that out, v2 needs to support python2. The suggested change above should remove the offending syntax

stgm:
I have adopted your suggestions verbatim. Note that the exception handler is not actually triggered if there are unavailable packages in requirements.txt (I would argue that this is the most obvious use case for reporting an exception).

stgm:
Ah, I think this was a misguided comment. The check contains a fresh requirements.txt which only contains the correct packages.

kzidane:
Thank you @stgm and @crossroads1112!

stgm:
Works like a charm. Nice!

PR #177, opened 2019-08-20
v3.0.8

crossroads1112:
Saw that someone was running into this issue in front. Previously this would throw an error:

```
check:
  run: ./foo
  stdin: 3.2
  stdout: 4
```


brianyu28:
Was literally about to message you asking about this, @crossroads1112, you read my mind.

PR #183, opened 2019-09-27
Small changes to allow using check50 to test programs with very large input

mjdv:
This commit makes the following two changes:

* The error message displayed by `check.stdout` on timeout now displays a message reflecting that a timeout was the cause of the failed check.
* `check.stdin` now has an optional parameter for a human-readable string, shown to the user as part of the check, instead of the actual string supplied. `check.stdout` already had such a parameter.

Additionally, this commit adds two unit tests to the collection of tests to ensure that these changes work as expected.

These two changes are made to allow the following two use cases:

1. (My current use case) We want to test whether a sorting function was implemented in O(n * log(n)) or in O(n^2). Experimentation suggests that any reasonable implementation of an O(n * log(n)) sorting algorithm in C will take less than a second to sort 500,000 integers. On the other hand, no ordinary implementation of an O(n^2) (average-case) sorting algorithm will sort 500,000 integers within 10 seconds. This means we can test the distinction using check50. This requires us to supply enormous inputs to the function. If these are displayed to the user, as `check.stdin` does before this commit, the checks become unusable. Furthermore, if the error for the time-out does not tell the user that it is *because* of the time-out, they cannot learn from the error.
2. (My upcoming use case) In some cases we may want to send human-unreadable input to a program, e.g. [a graph represented as a series of numbers](https://open.kattis.com/problems/shortestpath1). In such a situation, it can be much more useful to supply a human-readable description of the graph, such as "complete graph on 10 vertices" or "a cycle with 100 edges" rather than a series of numbers.

### DIFF
--- a/check50/_api.py
+++ b/check50/_api.py
@@ -163,12 +163,15 @@ class run:
         command = "bash -c {}".format(shlex.quote(command))
         self.process = pexpect.spawn(command, encoding="utf-8", echo=False, env=full_env)
 
-    def stdin(self, line, prompt=True, timeout=3):
+    def stdin(self, line, str_line=None, prompt=True, timeout=3):
         """
         Send line to stdin, optionally expect a prompt.
 
         :param line: line to be send to stdin
         :type line: str
+        :param str_line: what will be displayed as the delivered input, a human \
+                           readable form of ``line``
+        :type str_line: str
         :param prompt: boolean indicating whether a prompt is expected, if True absorbs \
                        all of stdout before inserting line into stdin and raises \
                        :class:`check50.Failure` if stdout is empty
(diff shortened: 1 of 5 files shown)


check50/_api.py
```diff
@@ -263,7 +269,7 @@ def stdout(self, output=None, str_output=None, regex=True, timeout=3):
                 result += self.process.after
             raise Mismatch(str_output, result.replace("\r\n", "\n"))
         except TIMEOUT:
-            raise Failure(_("did not find {}").format(_raw(str_output)))
+            raise Failure(_("did not find {} within {} seconds").format(_raw(str_output), timeout))
```
crossroads1112: @dmalan @brianyu28  Do we want to expose the timeout of each check to students directly? e.g. if students don't print anything for hello, they'll see `did not find "hello, world\n" within 3 seconds" ` rather than just `did not find "hello, world\n"`. Otherwise, this looks aok to me.

Jelleas: I can see the use case for a more advanced crowd to be able to actually see the timeouts. However, for problems like mario/caesar, I believe it's more confusing to see "within ... seconds" than not. As it might mistakenly cause you to believe that your program is too slow, whereas in reality your program is probably waiting for input erroneously.

Instead I added a `show_timeout` optional param to `.stdout` that lets you opt-in for the `within ... seconds` message.

Jelleas:
This has grown quite stale, but I still think it is a good idea. Lets merge @cmlsharp ?

PR #198, opened 2020-03-23
Show program output + monospace logs in webview

Jelleas:
Introducing `check50.Missing`. A custom Failure for when check50 is looking for an item, but it can't find it. 

The use case here is primarily the occurrence of a timeout in `.stdout()`. In which case check50 would've failed to find some string/regex in stdout. Currently check50 raises a generic Failure here, but that would truncate the expected output (other than in the log), as it needs to fit in a small terminal, and not show the output of the program up to that point. See example below:

image

`check50.Missing` addresses this and generates the below:

image

I've also taken the liberty of introducing a monospace font in the logs (through `<samp>`). As more often than not, there is some form of code/output in the logs that would otherwise be hard to read.

### DIFF
--- a/check50/_api.py
+++ b/check50/_api.py
@@ -263,7 +263,7 @@ class run:
                 result += self.process.after
             raise Mismatch(str_output, result.replace("\r\n", "\n"))
         except TIMEOUT:
-            raise Failure(_("did not find {}").format(_raw(str_output)))
+            raise Missing(str_output, self.process.before)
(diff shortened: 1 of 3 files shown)


Jelleas:
ps. I removed the failing check in https://github.com/cs50/check50/pull/199

I cannot replicate the failure locally and it seems weird that we're checking for a process to be alive after it's given us all of its output.

PR #199, opened 2020-03-23
--verbose debug

Jelleas:
Introducing the ability to set verboseness levels through command line options instead of environment variables (previously `CHECK50_LOGLEVEL`).

Default behaviour is unchanged, and `check50 -v` and `check50 --dev` behave as they did before. However, there's now the option to set verboseness through the `-v`/`--verbose` params:

```
check50 -v info
```

Is synonymous to:

```
check50 -v
```

And would, just like before, show all the git commands that lib50 runs and imply `--log`.

Now there's also:

```
check50 -v debug
```

That in addition to `-v info` sets the Python log level to `DEBUG`. This is where `lib50` logs all of its command output, and through this you get all this output in check50.

### DIFF
--- a/check50/__main__.py
+++ b/check50/__main__.py
@@ -240,7 +240,7 @@ def main():
     parser.add_argument("slug", help=_("prescribed identifier of work to check"))
     parser.add_argument("-d", "--dev",
                         action="store_true",
-                        help=_("run check50 in development mode (implies --offline and --verbose).\n"
+                        help=_("run check50 in development mode (implies --offline and --verbose info).\n"
                                "causes SLUG to be interpreted as a literal path to a checks package"))
     parser.add_argument("--offline",
                         action="store_true",
@@ -266,8 +266,14 @@ def main():
                         metavar="FILE",
                         help=_("file to write output to"))
     parser.add_argument("-v", "--verbose",
-                        action="store_true",
-                        help=_("display the full tracebacks of any errors (also implies --log)"))
+                        action="store",
+                        nargs="?",
+                        default="none",
+                        const="info",
+                        choices=["info", "debug"],
+                        help=_("sets the verbosity level and implies --log."
+                               " \"info\" displays the full tracebacks of errors and shows all commands run."
+                               " \"debug\" adds the output of all command run."))
     parser.add_argument("-V", "--version",
                         action="version",
                         version=f"%(prog)s {__version__}")
@@ -278,7 +284,6 @@ def main():
     global SLUG
     SLUG = args.slug
 
-
     if args.dev:
         args.offline = True
         args.verbose = True
@@ -286,9 +291,14 @@ def main():
     if args.offline:
         args.local = True
 
-    if args.verbose:
-        # Show lib50 commands being run in verbose mode
-        logging.basicConfig(level=os.environ.get("CHECK50_LOGLEVEL", "INFO"))
+    if args.verbose == "info":
+        # Show lib50 commands being run
+        logging.basicConfig(level=logging.INFO)
+        lib50.ProgressBar.DISABLED = True
+        args.log = True
+    elif args.verbose == "debug":
+        # Show lib50 commands being run and their output
+        logging.basicConfig(level=logging.DEBUG)
         lib50.ProgressBar.DISABLED = True
         args.log = True
 
--- a/check50/__main__.py
+++ b/check50/__main__.py
@@ -159,6 +159,31 @@ def compile_checks(checks, prompt=False):
     return "__init__.py"
 
 
+def setup_logging(level):
+    """
+    Sets up logging for lib50.
+    level 'info' logs all git commands run to stderr
+    level 'debug' logs all git commands and their output to stderr
+    """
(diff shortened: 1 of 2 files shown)


check50/__main__.py
```diff
@@ -266,8 +266,14 @@ def main():
                         metavar="FILE",
                         help=_("file to write output to"))
     parser.add_argument("-v", "--verbose",
-                        action="store_true",
-                        help=_("display the full tracebacks of any errors (also implies --log)"))
+                        action="store",
+                        nargs="?",
+                        default="none",
```
crossroads1112, line 271: Why not make the default something falsey (None/False) instead? That way things like "if args.verbose" would still work.

check50/__main__.py
```diff
@@ -278,17 +284,25 @@ def main():
     global SLUG
     SLUG = args.slug
 
-
     if args.dev:
         args.offline = True
-        args.verbose = True
+        args.verbose = "info"
```
kzidane, line 289: Should that override `args.verbose` only if `args.verbose` isn't passed? What if the user wants another verbosity level in dev mode?

check50/__main__.py
```diff
@@ -266,8 +266,14 @@ def main():
...
-                        action="store_true",
-                        help=_("display the full tracebacks of any errors (also implies --log)"))
+                        action="store",
+                        nargs="?",
+                        default="none",
+                        const="info",
+                        choices=["info", "debug"],
+                        help=_("sets the verbosity level and implies --log."
+                               " \"info\" displays the full tracebacks of errors and shows all commands run."
+                               " \"debug\" adds the output of all command run."))
```
kzidane, line 276: I would use single quotes here for the outer quotes instead of escaping double quotes in double quotes.

check50/__main__.py
```diff
@@ -278,17 +284,25 @@ def main():
...
         args.local = True
 
-    if args.verbose:
-        # Show lib50 commands being run in verbose mode
-        logging.basicConfig(level=os.environ.get("CHECK50_LOGLEVEL", "INFO"))
+    if args.verbose == "info":
+        # Show lib50 commands being run
+        lib50_logger = logging.getLogger("lib50")
+        lib50_logger.setLevel(logging.INFO)
+        lib50_logger.addHandler(logging.StreamHandler(sys.stdout))
```
kzidane, line 298: I wonder if log output should be logged out to `stderr` instead. That way you can easily do 

```
check50 ... 2>check50.log
```
To get a file only containing the logged output which might be useful?

check50/__main__.py
```diff
@@ -278,17 +284,25 @@ def main():
...
+        lib50.ProgressBar.DISABLED = True
+        args.log = True
+    elif args.verbose == "debug":
+        # Show lib50 commands being run and their output
+        lib50_logger = logging.getLogger("lib50")
+        lib50_logger.setLevel(logging.DEBUG)
+        lib50_logger.addHandler(logging.StreamHandler(sys.stdout))
```
kzidane, line 305: Should these lines be factored out into their own separate function to be able to reuse for `info` as well?

check50/__main__.py
kzidane: Agreed.

crossroads1112:
Should we do this for submit50 too? It also takes the loglevel from an env var (`SUBMIT50_LOGLEVEL`)

check50/__main__.py
Jelleas: You're right, done!

check50/__main__.py
Jelleas: I went with `""` instead of `False`/`None`, that at least keeps the type the same.

check50/__main__.py
Jelleas: Done!

check50/__main__.py
Jelleas: Done

check50/__main__.py
Jelleas: Arguably a lot more from `main` should be factored out into separate functions... But I factored this out now ;)

check50/__main__.py
```diff
@@ -278,42 +307,47 @@ def main():
...
     excepthook.output_file = args.output_file
 
+    # If remote, push files to GitHub and await results
     if not args.local:
         commit_hash = lib50.push("check50", SLUG, internal.CONFIG_LOADER, data={"check50": True})[1]
         with lib50.ProgressBar("Waiting for results") if "ansi" in args.output else nullcontext():
             tag_hash, results = await_results(commit_hash, SLUG)
+    # Otherwise run checks locally
     else:
-        with lib50.ProgressBar("Checking") if not args.verbose and "ansi" in args.output else nullcontext():
```
Jelleas, line 309: One minor inconsistency I noted was that we were always printing out the ProgressBars regardless of verbosity level in remote mode:

```
Connecting....
Authenticating...
...
```

But in `--local` or `--offline` mode with any verbosity level we would stop printing the `Checking...` ProgressBar. So I made that consistent. check50 now always prints the ProgressBars, but does not animate them (with more dots) with any set verbosity level.

Issue #200, opened 2020-04-06
break up --offline into --download-checks and --install-dependencies


kzidane:
When `--offline` is passed, `check50` doesn't install the `pip` packages specified as dependencies for checks. We currently use `--offline` to support private checks in remote `check50` to avoid having `check50` try to clone the checks. One way to workaround this is to add something like `--download-checks` and `--install-dependencies` as a mutually exclusive group with `--offline`, which would imply `--download-checks=false --install-dependencies=false` in this scenario. This should give us more granular control in remote `check50` such that we can avoid cloning the checks but still install dependencies.

crossroads1112:
Oh god, the number of "levels"  increases lol. I don't love the "=false" part though. Why not just boolean flags "--no-download-checks" and "--no-install-dependencies" instead since they're enabled by default? Similar to what e.g. GCC does for enabling/disabling warnings and errorrs

kzidane:
ikr? :sweat_smile: Not opposed to `--no-*`.

PR #202, opened 2020-04-13
--no-install-dependencies & --no-download-checks

Jelleas:
Fixes https://github.com/cs50/check50/issues/200

### DIFF
--- a/check50/__main__.py
+++ b/check50/__main__.py
@@ -336,7 +336,7 @@ def main():
 
             install_translations(config["translations"])
 
-            if args.no_install_dependencies:
+            if not args.no_install_dependencies:
                 install_dependencies(config["dependencies"], verbose=args.verbose)
 
             checks_file = (internal.check_dir / config["checks"]).resolve()
--- a/check50/__main__.py
+++ b/check50/__main__.py
@@ -299,6 +299,16 @@ def main():
         lib50.ProgressBar.DISABLED = True
         args.log = True
 
+    # Warning in case of running remotely with no_download_checks or no_install_dependencies set
+    if (args.no_download_checks or args.no_install_dependencies) and not args.local:
+        if args.no_download_checks and args.no_install_dependencies:
+            msg = _("--no-downloads-checks and --no-install-dependencies have")
+        elif args.no_download_checks:
+            msg = _("--no-downloads-checks has")
+        else:
+            msg = _("--no-install-dependencies has")
+        termcolor.cprint(_("Warning: {} no effect without --local").format(msg), "yellow", attrs=["bold"])
+
     # Filter out any duplicates from args.output
     seen_output = set()
     args.output = [output for output in args.output if not (output in seen_output or seen_output.add(output))]
--- a/check50/__main__.py
+++ b/check50/__main__.py
@@ -244,7 +244,7 @@ def main():
                                "causes SLUG to be interpreted as a literal path to a checks package"))
     parser.add_argument("--offline",
                         action="store_true",
-                        help=_("run checks completely offline (implies --local)"))
+                        help=_("run checks completely offline (implies --local, --no-download-checks and --no-install-dependencies)"))
     parser.add_argument("-l", "--local",
                         action="store_true",
                         help=_("run checks locally instead of uploading to cs50"))
@@ -268,6 +268,12 @@ def main():
     parser.add_argument("-v", "--verbose",
                         action="store_true",
                         help=_("display the full tracebacks of any errors (also implies --log)"))
+    parser.add_argument("--no-download-checks",
+                        action="store_true",
+                        help=_("do not download checks, but use previously downloaded checks instead (only works with --local)"))
+    parser.add_argument("--no-install-dependencies",
+                        action="store_true",
+                        help=_("do not install dependencies (only works with --local)"))
     parser.add_argument("-V", "--version",
                         action="version",
                         version=f"%(prog)s {__version__}")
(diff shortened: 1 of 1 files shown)


check50/__main__.py
```diff
@@ -268,6 +268,12 @@ def main():
     parser.add_argument("-v", "--verbose",
                         action="store_true",
                         help=_("display the full tracebacks of any errors (also implies --log)"))
+    parser.add_argument("--no-download-checks",
+                        action="store_true",
+                        help=_("do not download checks, but use previously downloaded checks instead (only works with --local)"))
```
kzidane, line 273: I assume this only works with `--local` because you don't wanna make sure that `.cs50.yaml` is up-to-date before pushing? Should this be enforced by the code? i.e., if you pass `--no-download-checks` without passing `--local`, should there be an error? Or should `--no-download-checks` imply `--local`?

Jelleas: It only works with `--local` because we'd otherwise have to put it in a commit message to carry it over to kareem.py. And I don't see any reason to support that, because at that point you should configure the checks differently. 

I don't like `--no-download-checks` implying `--local`. Conceptually this is more of a configuration option on top of `--local`, and not something that changes an entire mode of operation. The way I thought about it, even in `--remote` these options still make sense in that you will never locally download checks and/or install dependencies with these options set. It's just that in `--remote` these wouldn't be installed/downloaded in the first place ;)

kzidane: Right, I totally agree using `--no-download-checks` locally should have no impact on remote `check50`.

> I don't like --no-download-checks implying --local. Conceptually this is more of a configuration option on top of --local, and not something that changes an entire mode of operation. The way I thought about it, even in --remote these options still make sense in that you will never locally download checks and/or install dependencies with these options set. It's just that in --remote these wouldn't be installed/downloaded in the first place ;)

There's either a bit of a contradiction here or maybe I'm confused. So it "only" works with `--local` but it also works with `--remote` and it makes sense there?

I don't think it makes sense with `--remote` because if you specify it in this case, the `.cs50.yaml` file that you have locally could be different from that in the HEAD of the branch (the one that's ultimately going to be used by EKS since this option doesn't impact remote configs) which sounds weird. So for example, local `.cs50.yaml` might be excluding files that remote `.cs50.yaml` is requiring, and yet `check50` would still make a push without the files required by remote if that makes sense?

Jelleas: Ah I see the confusion! In `--remote` check50 does not actually clone the checks repo. Instead check50 fetches `.cs50.yml` directly from github.com. We do this for exactly the reason you described :). This is in lib50:

https://github.com/cs50/lib50/blob/f2fb3ae7c9ab0c6d40d924dc5d68be38ce2d500a/lib50/_api.py#L360

kzidane: Oh I see. But this still means that `--no-download-checks --no-install-dependencies` will do nothing if you run remote `check50`, right?

Jelleas: Yup, setting or unsetting those flags in `--remote` does nothing because what it touches isn't done in that mode anyway.

Jelleas: Perhaps we could change the phrasing from `only works in --local` to `only affects --local`?

kzidane: So they should either imply `--local` or error without it or else they could be a source of confusion. Doesn't it feel weird to you passing options that do nothing? :sweat_smile:

Jelleas: It feels weird, but I do think it's right given the alternatives. 

The `--local` implication option is going to cause even more confusion. It feels weirder to me that adding `--no-install-dependencies` changes the (default) mode of operation. Such that this:

````
check50 cs50/problems/2018/x/hello
````

Runs on a remote server, but this:

```
check50 cs50/problems/2018/x/hello --no-install-dependencies
```

Now runs on your machine because it's not installing dependencies?

Alternatively, we could error in the latter case. I'm not strongly opposed to this, because it is a useless flag after all. But, technically, what the flag tells you is still true. It is not installing dependencies (on your local machine). 

Perhaps then we should issue a warning? `"setting --no-install-dependencies has no effect without also setting --local"`

Comment (kzidane) [reply check50/__main__.py:273]: Fair. At this point, if we decide that `--no-*` imply `--local`, I would show a warning or something that says that `check50` will be running in local mode if `--local` is not explicitly specified or perhaps the easier approach would be just exit with an error saying that this is only allowed in local mode instead of just silently ignoring the flags. Maybe I'm over-complicating things though. Any thoughts @crossroads1112 @dmalan @brianyu28?

Comment (Jelleas) [reply check50/__main__.py:273]: Fyi, from Docker's docs: `--memory-swap is a modifier flag that only has meaning if --memory is also set`

https://docs.docker.com/config/containers/resource_constraints/#--memory-swap-details

Comment (kzidane) [reply check50/__main__.py:273]: Right, but that's enforced?

```
$ docker run -it --rm --memory-swap 100m alpine
docker: Error response from daemon: You should always set the Memory limit when using Memoryswap limit, see usage.
See 'docker run --help'.
```

Comment (crossroads1112) [reply check50/__main__.py:273]: +1 for no-install-dependencies not implying local but displaying a warning if passed without it.

Comment (Jelleas) [reply check50/__main__.py:273]: Went with the following warning:

```
Warning: --no-downloads-checks and --no-install-dependencies have no effect without --local
```

File: check50/__main__.py [hunk 2]
```diff
@@ -299,6 +299,16 @@ def main():
         lib50.ProgressBar.DISABLED = True
         args.log = True
 
+    # Warning in case of running remotely with no_download_checks or no_install_dependencies set
+    if (args.no_download_checks or args.no_install_dependencies) and not args.local:
+        if args.no_download_checks and args.no_install_dependencies:
+            msg = _("--no-downloads-checks and --no-install-dependencies have")
```
crossroads1112, line 305: Just FYI, sentence fragments like these tend to make translation difficult.

PR #203, opened 2020-04-28
use sys.executable instead of python3 to pip install

Jelleas:
Ran into this on a machine with two versions of Python installed. check50 was installed on a newer version of Python, whereas python3 referred to the older version.

### DIFF
--- a/check50/__main__.py
+++ b/check50/__main__.py
@@ -119,7 +119,7 @@ def install_dependencies(dependencies, verbose=False):
             for dependency in dependencies:
                 f.write(f"{dependency}\n")
 
-        pip = [sys.executable, "-m", "pip", "install", "-r", req_file]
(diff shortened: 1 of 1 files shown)


crossroads1112:
I have no problem with this, but I feel like this used to be `sys.executable` at some point and was changed. @kzidane do you remember this at all?

Jelleas:
I share your vague feeling that this was `sys.executable` at some point. But I think we changed this to `python3 -m pip` from just `pip install` at sigcse last year because we couldn't get check50 to run on someone's laptop (https://github.com/cs50/check50/commit/e614b4b8ec8b3696d5a4698ccd34a6717d1f6f87). And I think we only found out about `sys.executable` later that year, but I don't know why we didn't change this then.

kzidane:
> @kzidane do you remember this at all?

I don't, I'm sorry!

kzidane:
LGTM!

Issue #207, opened 2020-05-11
log levels, --log, --log-level, --verbose


kzidane:
I was gonna reply to @crossroads1112's comment https://github.com/cs50/check50/pull/206#issuecomment-625980981 but I thought maybe opening a new issue would be better since this PR is merged.

I actually thought we left `choices` as it was after all @Jelleas (i.e., `["debug", "info"]`) but maybe I was looking at a cached version. It seems there's no clean way to list the log level names unfortunately so maybe we should just hard-code their names. It's not like they change everyday anyway.  And good point, @crossroads1112 regarding `NOSET`. I guess that would be equivalent to a `--quiet` option that I don't think we have so perhaps we should add that. While we're at it, is there a reason why `--log` exists at all? I get that it lets you see the logs from the checks, but why is that not simply one of the log levels (e.g., `debug`) for example? Or perhaps `info` since it's lower than `debug` and then `debug` would show more `check50` logs.

And should we change `--verbose` to `--log-level` (and remove `--log` in case it just becomes another log level) to be consistent with other tools?

Jelleas:
The only thing `--log` does is show the log from each check in the ANSI output:

![image](https://user-images.githubusercontent.com/1257775/81542911-ab24e500-9375-11ea-9cf9-c729e27964ee.png)

I believe `--log` existed before the web UI, and was previously the only way to get access to these logs. However, now that the web UI and JSON output exist, and that we always show the log output in these (except in the case of `hidden` checks, but don't worry about that), you could argue that `--log` is a little redundant. Come to think of it, I don't think I've ever actually seen anyone use `--log`. On that note, I'm okay with cutting it altogether. @crossroads1112 thoughts?

`--verbose` to `--log-level`. Yes, if we drop `--log`.

"just hard-code their names" I hate everything. But you're probably right.

crossroads1112:
Can we get rid of NOTSET? It's literally the same as not passing the option at all, right?

crossroads1112:
Also, we need to change the error message in `check50.c._check_valgrind` because it says to rerun with `--log` to see the full valgrind output.

crossroads1112:
As for --log vs the others, I'm kinda torn. 

First, `--verbose` has more or less the same effect across every one of our tools, namely that it displays the full error message if an exception is thrown rather than the "Contact sysadmins" one. Now this is no longer the case per the latest change `--verbose` takes an argument so `check50` is now different from the other tools.

[...]

PR #208, opened 2020-05-11
mv --log --ansi-log + introduce --log-level

Jelleas:
https://github.com/cs50/check50/issues/207

Renames `--log` to  `--ansi-log` 

Introduces `--log-level` that can be set to `["DEBUG", "INFO", "WARNING", "ERROR", "CRITICAL"]`. These are the five unique values of the `logging` module as per https://docs.python.org/3/library/logging.html#levels. Technically `NOTSET`, `WARN` and `FATAL` also exist. However, `WARN` and `FATAL` are just synonyms for `WARNING` and `CRITICAL`. And `NOTSET` is realistically a confusing no op unless we introduce a custom logging level below `DEBUG`. 

`--verbose` is back to being a boolean that now, because of `--log-level`, only toggles whether we show tracebacks or not. Alternatively we could rename to `--traceback`?


crossroads1112:
Is it weird that we have a log-level literally called "WARNING" and we display a message saying "Warning: " when you use `--no-download-checks` in non-local mode, but these have nothing to do with one another?

crossroads1112:
Also, this implementation _always_ disables lib50.ProgressBar, which is probably bad.

Jelleas:
> Is it weird that we have a log-level literally called "WARNING" and we display a message saying "Warning: " when you use `--no-download-checks` in non-local mode, but these have nothing to do with one another?

Somewhat? Guess we might as well log it instead of printing it.

crossroads1112:
Okay, so I gave my take on logging in #209 (fixing most of these things, and also adding some additional validation)

crossroads1112:
To be clear, my PR was just a PR into this branch, we'd still need to PR into develop, so we can leave this open.

check50/__main__.py
```diff
@@ -385,9 +432,10 @@ def main():
             # Have lib50 decide which files to include
             included = lib50.files(config.get("files"))[0]
 
-            with open(os.devnull, "w") if args.verbose else nullcontext() as devnull:
-                # Redirect stdout to devnull if some verbosity level is set
-                if args.verbose:
+            # Redirect stdout to devnull if verbose or log level is set
+            should_redirect_devnull = args.verbose or args.log_level
+            with open(os.devnull, "w") if should_redirect_devnull else nullcontext() as devnull:
```
crossroads1112, line 437: Wait, was this even right in the first place? Shouldn't we redirect output to /dev/null if we _aren't_ in verbose mode?

Jelleas: Yup this is wrong.

crossroads1112:
Hopefully one last thought: is it weird that log level doesn't imply verbose? So like if I do log-level debug I'd see all this information but if an exception is thrown I still end up seeing "contact sysadmins" at the end?

Jelleas:
> Hopefully one last thought: is it weird that log level doesn't imply verbose? So like if I do log-level debug I'd see all this information but if an exception is thrown I still end up seeing "contact sysadmins" at the end?
> 
> I'm not sure what the right thing is to do here though since conceptually log-level error would be the appropriate level to show that info but our default level is warning.

I can see the conflict here, because we set the default at `WARNING` for purposes of showing the tools own warnings. However, the errors that we're hiding in non-verbose mode are not so much errors of the tool, but rather of individual checks that are ultimately student facing. So different audience, different place. On top of that the output of `--verbose` is part of the result of check50, it ends up in the actual `Check50Result` and this is visible even when running remotely.  

In that sense I don't think it's too weird to have two flags right now. One for configuring the logs, which is likely only useful for sysadmins in an environment where check50 is running in the background. because `--log-level` effectively lets you see what check50 does on your machine in more detail. And one (`--verbose`) that effectively lifts the "student-friendly" succinctness of check50.

Altneratively, we could introduce our own logging level in between `INFO` and `WARNING`. Something like `TRACEBACK` perhaps? We could then log any tracebacks to this level instead. However, we would then still need to figure out a way to get any tracebacks the remote end back to the local run of check50 as well.

Jelleas:
> Hopefully one last thought: is it weird that log level doesn't imply verbose? So like if I do log-level debug I'd see all this information but if an exception is thrown I still end up seeing "contact sysadmins" at the end?
> 
> I'm not sure what the right thing is to do here though since conceptually log-level error would be the appropriate level to show that info but our default level is warning.

Hmmmm, what `--verbose` currently does is a little weird. It prints any exception out, just before exiting check50, and does NOT redirect stdout/err to dev/null. 

This means there are some differences in running remote/local with `--verbose`. For instance in case of a syntax error (or any global error) in checks:

```
$ check50 jelleas/foo/master --local -v
Sorry, something's wrong! Let sysadmins@cs50.harvard.edu know!
Traceback (most recent call last):
  File "/usr/local/bin/check50", line 11, in <module>
    load_entry_point('check50==3.0.10', 'console_scripts', 'check50')()
  File "/usr/local/lib/python3.7/site-packages/check50/__main__.py", line 354, in main
    runner = CheckRunner(checks_file)
  File "/usr/local/lib/python3.7/site-packages/check50/runner.py", line 176, in __init__
    self.checks_spec.loader.exec_module(check_module)
  File "<frozen importlib._bootstrap_external>", line 724, in exec_module
  File "<frozen importlib._bootstrap_external>", line 860, in get_code
  File "<frozen importlib._bootstrap_external>", line 791, in source_to_code
  File "<frozen importlib._bootstrap>", line 219, in _call_with_frames_removed
  File "/root/.local/share/check50/jelleas/foo/__init__.py", line 3
    this is invalid syntax
                         ^
SyntaxError: invalid syntax
```

versus:

```
$ check50 jelleas/foo/master -v
check50 ran into an error while running checks! Please contact sysadmins@cs50.harvard.edu!
Traceback (most recent call last):
  File "/usr/local/bin/check50", line 11, in <module>
    load_entry_point('check50==3.0.10', 'console_scripts', 'check50')()
  File "/usr/local/lib/python3.7/site-packages/check50/__main__.py", line 307, in main
    tag_hash, results = await_results(commit_hash, SLUG)
  File "/usr/local/lib/python3.7/site-packages/check50/__main__.py", line 191, in await_results
    raise RemoteCheckError(results["check50"])
check50.__main__.RemoteCheckError: check50 ran into an error while running checks! Please contact sysadmins@cs50.harvard.edu!
Exception payload:
{"remote_json": {"error": {"data": {}, "traceback": ["  File \"/usr/local/bin/check50\", line 11, in <module>\n    load_entry_point('check50==3.0.10', 'console_scripts', 'check50')()\n", "  File \"/usr/local/lib/python3.7/site-packages/check50/__main__.py\", line 354, in main\n    runner = CheckRunner(checks_file)\n", "  File \"/usr/local/lib/python3.7/site-packages/check50/runner.py\", line 176, in __init__\n    self.checks_spec.loader.exec_module(check_module)\n", "  File \"<frozen importlib._bootstrap_external>\", line 724, in exec_module\n", "  File \"<frozen importlib._bootstrap_external>\", line 860, in get_code\n", "  File \"<frozen importlib._bootstrap_external>\", line 791, in source_to_code\n", "  File \"<frozen importlib._bootstrap>\", line 219, in _call_with_frames_removed\n"], "type": "SyntaxError", "value": "invalid syntax (__init__.py, line 3)"}, "slug": "jelleas/foo/master", "version": "3.0.10"}}
```

This badly formatted output happens because we raise yet another error (`RemoteCheckError`) in case we find an `"error"` attribute in the remote's json.

Another weird thing, because we do not redirect stdout/err to /dev/null in  `--verbose`, is that you see any print messages locally, but they disappear when running remotely. Sure, makes sense once you think it all through, but it is pretty weird if you would try to add a print statement to a check (by editing directly on GH perhaps) to then find out that you don't actually see that print statement even with `--verbose`. 

Finally,  to my surprise, `--verbose` has no effect on tracebacks in checks. We always show these. Why do we do this?

My suggestion then, given what `--verbose` does, how about we introduce our own logging level `TRACEBACK` that sits in between `INFO` and `WARNING`, and will show you the Tracebacks of any exceptions. Including those that get send back from check50's remote end. And let's hide any Tracebacks by default even those from checks?

We then scrap `--verbose` all together, and move the ability to see stdout/err to `--dev`. Because `--dev` implies `--local`, and it is conceptually probably the only time you should want to see stdout/err from your checks anyway.

check50/__main__.py
```diff
@@ -285,15 +361,20 @@ def main():
...
     parser.add_argument("-v", "--verbose",
+                        action="store_true",
+                        help=_("shows the full traceback of any errors and shows print statements written in checks"))
+    parser.add_argument("--log-level",
                         action="store",
-                        nargs="?",
-                        default="",
-                        const="info",
-                        choices=[attr for attr in dir(logging) if attr.isupper() and isinstance(getattr(logging, attr), int)],
+                        choices=list(LogLevel.__members__),
```
kzidane, line 368: Instead of accessing `__members__`, something like the following is probably cleaner:

```
[level.name.lower() for level in LogLevel]
```

kzidane:
> My suggestion then, given what --verbose does, how about we introduce our own logging level TRACEBACK that sits in between INFO and WARNING

I would say we shouldn't introduce any special log levels unless we absolutely have to just for simplicity and to stay as close to the standards as possible.

> We always show these. Why do we do this?

IMO we should always show tracebacks instead of just showing "contact sysadmins@" because when they contact sysadmins@, the first thing we ask users to do is re-run again, with different options, to get some more output to see what's happening. By always showing tracebacks, we can potentially save a step.

> We then scrap --verbose all together, and move the ability to see stdout/err to --dev. Because --dev implies --local, and it is conceptually probably the only time you should want to see stdout/err from your checks anyway.

I personally think `--verbose` is unnecessary given `--log-level`. Sure, it saves you a few keystrokes, but it honestly feels redundant to me.

check50/__main__.py
```diff
@@ -25,9 +27,36 @@
...
+class ColoredFormatter(logging.Formatter):
+    COLORS = {
+        "ERROR": "red",
+        "WARNING": "yellow",
+        "INFO": "cyan",
+        "DEBUG": "magenta",
+    }
+
+    def __init__(self, fmt, use_color=True):
+        logging.Formatter.__init__(self, fmt=fmt)
```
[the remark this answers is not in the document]
crossroads1112: Yup

check50/__main__.py
```diff
@@ -253,28 +281,76 @@ def raise_invalid_slug(slug, offline=False):
     raise internal.Error(msg)
 
 
+def validate_args(args):
```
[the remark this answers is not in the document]
crossroads1112: I mean it isn't parsing args though. That's already done. I agree it does more than validate but I don't think parse is right.

check50/__main__.py
```diff
@@ -253,28 +281,76 @@ def raise_invalid_slug(slug, offline=False):
...
+            args.ansi_output = True
+
+    # offline implies local
+    if args.offline:
+        args.no_install_dependencies = True
+        args.no_download_checks = True
+        args.local = True
+
+
+    args.log_level = LogLevel.__members__[args.log_level.upper()]
```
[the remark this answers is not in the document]
crossroads1112: I forgot enums implemented `__getitem__`. yeah that's fine. I don't think we really need to check for KeyError since thats already done by argparse as you said. I'm okay with relying on that invariant, but there's no harm in adding it I guess.

crossroads1112:
Agreed re: always printing tracebacks.

check50/__main__.py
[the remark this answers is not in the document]
Jelleas: I went with `process_args` for now. It's vague, but at least it hints at doing more than just validation

Jelleas:
> always printing tracebacks

There are a couple of situations where we don't want to print tracebacks, just to name a couple: an invalid slug, a github error or a config error. In those cases it'd just be confusing to show where the exception came from. For this check50 used to, and that's probably right, hide the tracebacks of any check50 or lib50 error. 

But I do agree, we should show tracebacks by default (and instruct students to email those to sysadmins@) for any other error.

**Here's the catch**, there's two ways an error can occur. Either locally, in which case check50 has the type (class) of the exception and can deduce via the standard way of subclasses on what to do. Or remotely, in which case check50 currently communicates the type of the error back as a string. But there's no way to go from that string (or any other string) to a class. Not only because of Python's import system, but also because that type of error might have come from an extension that is not available locally. This makes it impossible to rely on subclassing to deduce what to do with a remote error. 

To solve this I've added a field to check50's error json: `"actions"` (open to alternative names) that stores what to do with the exception. The idea being that on the remote end, check50 can still figure out through subclasses whether the traceback should be printed and what the message to the student should be. And can then send that back to where the student ran check50, which can then display the traceback/message just as if the student ran with `--local` themselves.

Here's what the json looks like:

```json
json = {
    "slug": "jelleas/foo/master",
    "error": {
        "type": "InvalidSlugError",
        "value": "foo",
        "traceback": [
            "Traceback (most recent call last):\n",
            "bar\n"
        ],
        "actions": {
            "show_traceback": true,
            "message": "foo"
        },
        "data": {}
    },
    "version": "3.1.1"
}```

**Comment by cmlsharp** [comment]
What's the difference between `.error.value` and `.error.actions.message`? When do these differ? Do we need both?

File: check50/internal.py [hunk 6]
```diff
@@ -35,6 +35,13 @@ class Error(Exception):
     pass
 
 
+class RemoteCheckError(Error):
```
cmlsharp, line 38: Should this be in `internal`? internal is traditionally the place for the extension-writer API. `RemoteCheckError` isn't part of that API.

cmlsharp:
Also, what does verbose do now? The help message still says that it prints tracebacks.

check50/internal.py
Jelleas: I agree, but I don't think we have a better place to put it right now. Perhaps we should introduce `exceptions.py`? And move `Error`, `RemoteCheckError` and `_excepthook` to that instead?

Jelleas:
> Also, what does verbose do now? The help message still says that it prints tracebacks.

Oh looks like that get overwritten in the merge. Verbose still does two things: it lets you see any print statements in checks if running locally, and it shows which dependencies get installed.

check50/internal.py
Jelleas: inb4 `truly_internal_dont_touch.py`

cmlsharp:
I added exceptions.py. Thoughts?

check50/__main__.py
```diff
@@ -118,20 +122,19 @@ def setup_logging(level):
...
 
-    # Direct all logs to sys.stderr
-    lib50_logger.addHandler(logging.StreamHandler(sys.stderr))
+        # Direct all logs to sys.stderr
+        logger.addHandler(handler)
 
-    # Don't animate the progressbar
-    lib50.ProgressBar.DISABLED = True
+    # Don't animate the progressbar if LogLevel is either info or debug
+    lib50.ProgressBar.DISABLED = logger.level < LogLevel.WARNING
```
kzidane, line 137: Does that still work as expected with `WARNING` now set to `auto()`, which I think is `2` in this case?

check50/__main__.py
```diff
@@ -203,7 +206,55 @@ def raise_invalid_slug(slug, offline=False):
...
+
+    # Filter out any duplicates from args.output
+    seen_output = []
+    for output in args.output:
+        if output in seen_output:
+            LOGGER.warning(_("Duplicate output format specified: {}").format(output))
+        else:
+            seen_output.append(output)
+
+    args.output = seen_output
```
kzidane, line 253: Should we just use a `set` out of `args.output` and use it? We could still show a generic warning that duplicate output formats are ignored if the size of the set is less than the size of the list if we want to but I don't think it's a big deal.

check50/__main__.py
```diff
@@ -238,15 +286,20 @@ def main():
...
+                        default="warning",
+                        choices=[level.name.lower() for level in LogLevel],
+                        type=str.lower,
+                        help=_("sets the log level."
+                               ' "warning" displays usage warnings'
+                               ' "info" shows all commands run.'
+                               ' "debug" adds the output of all command run.'))
+    parser.add_argument("--ansi-log",
+                        action="store_true",
+                        help=_("display log in ansi output mode"))
```
kzidane, line 302: Why is that separate from `--log-level` again?

check50/__main__.py
```diff
@@ -203,7 +206,55 @@ def raise_invalid_slug(slug, offline=False):
...
+    for output in args.output:
+        if output in seen_output:
+            LOGGER.warning(_("Duplicate output format specified: {}").format(output))
+        else:
+            seen_output.append(output)
+
+    args.output = seen_output
+
+    if args.ansi_log and "ansi" not in seen_output:
+        LOGGER.warning(_("--ansi-log has no effect when ansi is not among the output formats"))
```
kzidane, line 256: I honestly still don't like these useless arguments and would much rather exit with an error code if an argument that's useless or shouldn't be used is passed.

check50/__main__.py
Jelleas: Whoops good catch! Using `logging.INFO`, `logging.ERROR` etc. now instead of `enum.auto()`

check50/__main__.py
Jelleas: `--ansi-log` shows the student facing logs in the ansi output of the checks:

```
running clang...
looking for output...
```

`--log-level` controls "developer" (check writers, extension writers, Kareem's of this world) logs.

check50/__main__.py
Jelleas: I disagree, I think a warning is in place here because these command-line options are harmless. Misuse (useless use) of them does not change your environment, submit code that you wouldn't want to submit or anything of the sorts.

My thinking is, we should only switch to errors in case misuse of a command line option can have an unintended consequence.

check50/__main__.py
Jelleas: The original thinking here was that order mattered. If you supplied: `-o ansi json html`, check50 would then deliver output in that order. That's why it's implemented in the way it is.

But, I might go back on this because I'm not entirely sure this level of control is at all needed. For now, no reason to change though.

Jelleas:
> I added exceptions.py. Thoughts?

Looks good! Good thinking to re-use the callable class workaround for Excepthook.

check50/__main__.py
kzidane: It just seems to me that it's pretty much always better to enforce the rules. When an argument is either misused, may have unintended behaviors, or just doesn't make sense in a certain context it just seems better exit with an error code and force the user to remove it because there's absolutely no reason to let the user pass it when it's not useful and not needed. Just like that Docker CLI example that you brought up and it turned out they were enforcing it. Can we please change this behavior? :sob:

PR #209, opened 2020-05-13
add check50 logger, colored logging

crossroads1112:
Summary of changes: 

- Added our own `LogLevel` enum which defines the log levels we actually care about
- Add a `ColoredFormatter` which allows messages at levels to be displayed using different colors
- Show log level as part of the message
- Add check50 logger that is independent of lib50 logger. Levels are set the same way.
- Removed global SLUG variable in favor of excepthook.slug (this was inconsistent with what we do with every other variable we need the excepthook to know about).
- Moved argument validation into its own function because it's getting kinda out of hand
- Added `--ansi-log` (we can change the name as needed)
- Added additional warnings for duplicate output formats and specifying `--ansi-log` when ansi is not one of the output formats.
- Renamed `log` argument to `renderer.to_ansi` to `_log`. The current precedent is that for each top level key in the json there is a corresponding argument to each renderer. `log` is not a top-level key and if we ever added it in the future, it would be an issue for the ansi renderer. As such, I propose namespacing additional parameters to renderers via `_`. This is what flask does for `url_for` for example.


Jelleas:
The `LogLevel` enum is the inverse of the `logging` levels. This makes the following a little weird:

`if level > LogLevel.WARNING:`

As this reads as higher than WARNING, which in case of `logging` means ERROR and CRITICAL.

Perhaps change the enum to:

```
 class LogLevel(enum.IntEnum):
    ERROR = logging.ERROR
    WARNING = logging.WARNING
    INFO = logging.INFO
    DEBUG = logging.DEBUG
```

And then change:

`if level > LogLevel.WARNING:` => `if level <= LogLevel.INFO:`

Jelleas:
in `--dev` you now set log_level to be at least INFO via:

`args.log_level = max(args.log_level, LogLevel.INFO)` 

But that seems weird right? You'd want to be able to overwrite that?

Jelleas:
With the default now set `default="WARNING"`, there's no need for:

```
if not args.log_level:
        args.log_level = LogLevel.WARNING
```

crossroads1112:
@Jelleas Now that I think of it, should --dev actually imply `log-level info`? That's a departure from current behavior that may scare teachers who are developing tests. I feel like it should just imply `--ansi-log`

Jelleas:
https://github.com/cs50/check50/pull/209#issuecomment-628272414 agree!

PR #211, opened 2020-05-28
Log limit

Jelleas:
With clang potentially spewing out a 1000+ lines of output as per https://github.com/cs50/submit.cs50.io/issues/126 it seemed reasonable to impose a limit on logs. This PR limits the number of log lines from clang to a default of 50, and the total number of lines in the log to a default of 100. Both are configurable through a kwarg `max_log_lines`.

While I was at it, I limited the `max height` of the log in the webview and made it scroll on overflow.

### DIFF
--- a/check50/c.py
+++ b/check50/c.py
@@ -14,7 +14,7 @@ CC = "clang"
 CFLAGS = {"std": "c11", "ggdb": True, "lm": True}
 
 
-def compile(*files, exe_name=None, cc=CC, **cflags):
+def compile(*files, exe_name=None, cc=CC, max_log_lines=50, **cflags):
     """
     Compile C source files.
 
@@ -62,7 +62,7 @@ def compile(*files, exe_name=None, cc=CC, **cflags):
 
     # Log the last 50 lines of output in case compilation fails
     if process.exitcode != 0:
-        for line in stdout.splitlines()[-50:]:
+        for line in stdout.splitlines()[-max_log_lines:]:
             log(line)
         raise Failure("code failed to compile")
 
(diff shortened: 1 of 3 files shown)


check50/c.py
```diff
@@ -60,8 +60,9 @@ def compile(*files, exe_name=None, cc=CC, **cflags):
     # Strip out ANSI codes
     stdout = re.sub(r"\x1B\[[0-?]*[ -/]*[@-~]", "",  process.stdout())
 
+    # Log the last max_log_lines lines of output in case compilation fails
     if process.exitcode != 0:
-        for line in stdout.splitlines():
+        for line in stdout.splitlines()[-max_log_lines:]:
```
kzidane, line 65: Hmm, does it make more sense to show the first n lines instead of the last n? Should we limit the number of chars (instead of the number of lines) to control the size more?

kzidane: Should we consider using `-ferror-limit` too? https://clang.llvm.org/docs/UsersManual.html#cmdoption-ferror-limit

Jelleas: I think the last n lines make more sense, as that more likely shows what caused the check to fail and the information just before it is probably more relevant. Imagine seeing a traceback from Python perhaps, and not the actual error in the end ;)

`-ferror-limit` wouldn't help, because it's the warnings that blow up the results. Just checked again, and there are indeed just 20 errors shown (default value of `-ferror-limit`), but a good 629 warnings. Sadly I can't find `-fwarning-limit` or anything of the sorts for clang, but perhaps I overlooked something.

kzidane: I wasn't thinking about Python. For C, I think the first few lines are more relevant for sure, but I can see why it'd be useful to see the last few lines with Python. As a middle ground, should we just show 

```
the first n/2 lines 
... 
the last n/2 lines
```

?

Jelleas: Deal

PR #218, opened 2020-06-22
Tests update

jsarchibald:
This adds two test cases -- one to `check50_tests`, for when a Python file exists but it returns an unexpected error code; one to `c_tests`, for when a C file compiles incorrectly.


kzidane:
Thank you!

Issue #223, opened 2020-06-30
Can't touch check50.internal.check_dir on Mac with Python 3.8


Jelleas:
In short, Python 3.8 switched to `spawn` for multiprocessing:

> Changed in version 3.8: On macOS, the spawn start method is now the default. The fork start method should be considered unsafe as it can lead to crashes of the subprocess. [src](https://docs.python.org/3/library/multiprocessing.html#contexts-and-start-methods)

That means no memory is shared, and everything that is only initialized in the main process is missing from the child process (the check). And so is `check50.internal.check_dir`, but also `check50.__main__.excepthook.verbose`, `check50.__main__.excepthook.outputs` and `check50.__main__.excepthook.output_file`.

Jelleas:
Because this list is relatively short, I propose we add the following arguments to `check50.runner.CheckRunner.run` and `check50.runner.run_check`, and initialize all variables again in the check's process:

* `check_dir`
* `excepthook_settings`

Where `excepthook_settings` is just a tuple of tuples (something pickle-able) containing the name and actual setting:

`(("verbose", True), ("outputs", ("ansi", "json")), ("output_file", "foo"))`

PR #224, opened 2020-07-02
Allow for "spawn"

Jelleas:
Python 3.8 uses spawn as default on MacOS (https://github.com/cs50/check50/issues/223). From the [Python docs](https://docs.python.org/3/library/multiprocessing.html#contexts-and-start-methods):

> The fork start method should be considered unsafe as it can lead to crashes of the subprocess. See [bpo-33725](https://bugs.python.org/issue33725).

This PR re-instantiates all the module attributes that should be set in each check process, and through that check50 now supports spawn. To do this `runner.run_check` keeps a tuple `attribute_names`. Upon initialization of `run_check` all attribute names are evaluated (`eval()`) and stored with `run_check`. Then upon running `run_check` in the check process each attribute gets set (`setattr()`) to the stored  value from the parent process.

### DIFF
--- a/check50/runner.py
+++ b/check50/runner.py
@@ -291,7 +291,7 @@ class CheckRunner:
 
 class run_check:
     """
-    Check job that runs in the a child process.
+    Check job that runs in a seperate process.
     This is only a class to get around the fact that `pickle` can't serialize closures.
     This class is essentially a function that reimports the check module and runs the check.
     """
--- a/check50/runner.py
+++ b/check50/runner.py
@@ -290,7 +290,8 @@ class CheckRunner:
 
 class run_check:
     """
-    Hack to get around the fact that `pickle` can't serialize closures.
+    Check job that runs in the a child process.
+    This is only a class to get around the fact that `pickle` can't serialize closures.
     This class is essentially a function that reimports the check module and runs the check.
     """
 
--- a/check50/runner.py
+++ b/check50/runner.py
@@ -291,7 +291,7 @@ class CheckRunner:
 
 class run_check:
     """
-    Check job that runs in a seperate process.
+    Check job that runs in a separate process.
     This is only a class to get around the fact that `pickle` can't serialize closures.
     This class is essentially a function that reimports the check module and runs the check.
     """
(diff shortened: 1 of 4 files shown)


crossroads1112:
`internal.excepthook` should probably by `internal._excepthook` seeing as how we really don't want to let extension writers overwrite it. Obviously this doesn't prevent them from doing so, but it does make it clear that it is private.

check50/runner.py
```diff
@@ -299,8 +301,30 @@ def __init__(self, check_name, spec, checks_root, state=None):
         self.spec = spec
         self.checks_root = checks_root
         self.state = state
+        self.attribute_names = (
```
crossroads1112, line 304: why not make this a frozendict?

check50/runner.py
```diff
@@ -299,8 +301,30 @@ def __init__(self, check_name, spec, checks_root, state=None):
...
+            "internal.check_dir",
+            "internal.slug",
+            "internal.excepthook.outputs",
+            "internal.excepthook.output_file",
+            "internal.excepthook.verbose"
```
crossroads1112, line 309: Do we also need to include anything in 'internal' that would potentially be used by extensions?

Jelleas: Currently in our model, extensions only come into play when the checks' module is loaded. That happens inside the `CheckRunner` and that's currently unchangeable for extensions.  So the rationale here is, all that needs to be ported over to the individual check processes is in our control, and those things just so happen to live in `internal`.  

That said, if this does grows much further than this, or extension writers do require access, this should probably change to something similar in spirit to flask's `g`. Where if something needs to persist across processes, you assign it to some namespace / global store, and then `run_check` just copies from that namespace / store. 

But for now, let's KISS

check50/runner.py
Jelleas: Yeah why not

check50/runner.py
Jelleas: To future proof this a bit more, let me go ahead and change `attribute_names` from an instance variable to a class variable named `CROSS_PROCESS_ATTRIBUTES`. This should make it possible for others to change this list (okay, technically tuple).

Issue #222, opened 2020-06-28
Feature request: dynamically adding checks during execution?


pazz:
Hello once more, and sorry for spamming you with feature requests without contributing much first.

I am teaching an undergraduate "Intro to OOP" class, with exercises in Java. In order to facilitate check50 checks for structural inspection (class hierarchy, or method signatures), I wrote a small extension that allows to run junit5 unit tests from check50. At this point, its use is rather cumbersome due to the fact that one has to manually add check50 checks for every unit test on its own unless one is fine with one big check for all unit tests in a test class.

It'd be great if I could dynamically add checks to the dependency tree at execution time.

This would allow to first compile a (junit) test class, then extract the contained unit tests and interpret them as individual check50 checks for marking. In fact, I can easily run all unit test in one go and retroactively interpret the outcome. So it would even suffice to make check50 aware of new checks after the results are computed. But for the sake of consistent student feedback Id'be be great to at least make it look as if regular checks have been run. 

It looks to me as if check50's runner class computes the dependency DAG of checks before executing them.
Any pointers would be much appreciated.

Thanks again for an awesome tool,
P

Jelleas:
I might have something to help you here. You're correct in that check50 creates a dependency DAG prior to executing checks. And because checks run in a separate process, it's impossible, or at least very difficult, to change the DAG from within a check. However, to create this DAG, check50 will import the checks module. That means any code that is in that module will run. You can use this to dynamically create checks prior to running checks.

In fact I've done so in an extension of my own at: 

https://github.com/uva/ci_checks/blob/ace8dc6d45b69b7dcafa399c96d6e28c568596a3/mod1/__init__.py#L67-L95

Here I use one check (`exists`) to run all unittests in a Jupyter Notebook, but this might as well be JUnit. Then `exists` passes down the results in a large tuple. This can be any data structure that is serializable (can be pickled).

```Py
@check50.check()
def exists():
    ...
    for cell in cells:
        # Execute the cell
        try:
            execute(cell)
            exception = ""
            passed = True
        except check50.Failure as f:
            exception = str(f)
            passed = False

        # If it's a test cell, record result
        if check_jupyter.is_test_cell(cell):
            results.append((check_jupyter.get_cell_id(cell), passed, exception))

    # Pass down the results to the actual checks
    return tuple(results)
```

Then I dynamically create checks, that each depend on `exists` to receive its results. 

```Py
for test_id in get_test_ids("module 1.ipynb"):
        check = create_check(test_id)

[...]

PR #225, opened 2020-07-06
check50.internal.run_root_dir

Jelleas:
This PR introduces `check50.internal.run_root_dir` (effectively `check50.internal.run_dir.parent` while running a check) and `cd`s to that directory before importing the checks module.

This addresses two concerns:

1. Currently check50 does the first import of the checks module in the directory it was called from. That means that the following:

`__init__.py`
```Py
open("foo", "w")
```

Will create a file called `foo` in what is likely the directory the student is working from. Which is both unexpected and unwanted behavior.

2. Creates a persistent location for dynamically created files that need to exist across checks. For example in https://github.com/cs50/check50/issues/222#issuecomment-654267661 I used `check50.internal.run_root_dir` to store a list  of which checks need to exist, and each process then recreates the checks from that list.


Jelleas:
`CheckRunner`  changed to a contextmanager that uses `lib50.working_area` and `lib50.cd` underneath the hood. This cleans up the code `__main__.py`  and makes  `CheckRunner.__init__` inexpensive to run.

Jelleas:
Added `check50.internal.student_dir` that keeps track of the current working directory from where check50 was called. This makes it possible to access student files prior to running checks.

check50/internal.py
```diff
@@ -13,12 +13,19 @@
 #: Directory containing the check and its associated files
 check_dir = None
 
-#: Temporary directory in which check is being run
+#: Temporary directory in which the current check is being run
 run_dir = None
 
+#: Temporary directory that is the root (parent) of all run_dir(s)
```
crossroads1112, line 19: Why is this necessary instead of just using run_dir.parent?

Jelleas: `run_dir` only gets set inside a check, that is after having imported the checks module. Whereas `run_root_dir` gets set before importing. This makes it possible to store things prior to running checks.

PR #226, opened 2020-07-10
int and float support for run.stdout

Jelleas:
This adds int/float support to `run.stdout`. Whenever an int/float is passed as `output`, the following regex gets used to match that int or float exactly:

```Py
fr"(?<!\d){re.escape(str(output))}(?!\d|\.)"
```

drawing

This is the transformation, but do note that str_output does not get overwritten if set by the user:

```
.stdout(1)   => .stdout(r"(?<!\d)1(?!\d|\.)", str_output=str(1), regex=True)
.stdout(1.0) => .stdout(r"(?<!\d)1.0(?!\d|\.)", str_output=str(1.0), regex=True)
```

The hope is that this PR can eliminate the need for something like this in checks when dealing with numbers:

https://github.com/cs50/problems/blob/d2bad69a5e842cd34a179e6b6e36b5f3de5027ac/cash/__init__.py#L78-L80

One final note, a float with additional trailing zeros does **not** match:

```Py
print("1.00")
run.stdout(1.0)
```

My thinking here is that most of the time you'd want an exact match, and trailing zeros can still be supported by passing the regex yourself.

### DIFF
--- a/check50/_api.py
+++ b/check50/_api.py
@@ -1,5 +1,6 @@
 import hashlib
 import functools
+import numbers
 import os
 import re
 import shlex
@@ -136,6 +137,29 @@ def import_checks(path):
     return mod
 
 
+def number_regex(number):
+    """
+    Create a regular expression to match the number exactly:
+        (?<!\d)number(?!\d|(\.\d))
+
+    (?<!\d) = negative behind, \
+        asserts that there are no digits in front of the number.
+    (?!\d|(\.\d)) = negative lookahead, \
+        asserts that there are no digits and no . followed by digits after the number.
+
+    :param number: the number to match in the regex
+    :type number: any numbers.Number (such as int, float, ...)
+    :rtype: str
+
+    Example usage::
+
+        # Check that 7.0000 is printed with 5 significant figures
+        check50.run("./prog").stdout(check50.number_regex("7.0000"))
+
+    """
+    return fr"(?<!\d){re.escape(str(number))}(?!\d|(\.\d))"
+
+
 class run:
     """
     Run a command.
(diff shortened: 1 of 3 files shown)


Jelleas:
TIL `numbers.Number`

> Also, should this affect yaml checks? It could in principle since yaml makes distinctions between types. But maybe that's more complicated than we necessarily want.

We currently do the following for stdout in YAML checks:

```
    if isinstance(arg, list):
        arg = r"\n".join(str(a) for a in arg)
    arg = str(arg).replace("\n", r"\n").replace("\t", r"\t").replace('"', '\"')
    return f'.stdout("{arg}", regex=False)'
```

Here in case of a list, we join each entry with `\n` and do an exact match against it. That's somewhat incompatible with looking for numbers through a regex. It'd be possible with a more complex regex, but that complexity might scare some users away from the generated Python checks. So I vote we just keep the YAML checks the way they are.

> which then people could use with check50.run("./prog").stdout(check50.run.number_regex("1.00")) or something.

Fair enough. I made it part of `check50._api` instead of `run` because `run` only has methods that interact with the running process, and its methods are always chained. 

> should we have the |\. part at the end of the regex? 1. is the same as 1 and we also risk false positives if people put the number at the end of a sentence. I think we're better off without it.

Good catch there! We do need `.` in the negative lookahead somewhere, as the regex would otherwise accept a floating point number like `1.5`. Instead I changed the final part to the following that ignores a trailing `.`, but not if it's followed by a digit:

```
(?!(\.?\d))
```

cmlsharp:
Cool. Two things to consider:

Now that I think about it `decimal_regex` is probably a better name, since our regex only applies to base-10 numbers. We could generalize in principle via an additional `base` argument and keep the name `number`, but probably no reason to add complexity

Secondly, It may be possible in the future that more of these generally useful regex helpers come up. One way to future-proof this would be to have a namespace for these e.g. `check50.regex` so this would become `check50.regex.decimal`

Also, speaking of `number` regardless of whether we do that last thing,

cmlsharp:
Oh, also, what about negative numbers?

check50/_api.py
```diff
@@ -137,33 +137,33 @@ def import_checks(path):
...
-
-    (?<![\d\-]) = negative lookbehind, \
-        asserts that there are no digits and no - in front of the number.
-    (?!(\.?\d)) = negative lookahead, \
-        asserts that there are no digits and no additional . followed by digits after the number.
-
-    :param number: the number to match in the regex
-    :type number: any numbers.Number (such as int, float, ...)
-    :rtype: str
+class regex:
```
cmlsharp, line 140: I feel like this should just be a module rather than a class. A class that is just a collection of staticmethods is essentially the same as a module, but much less idiomatic.

PR #229, opened 2020-07-14
Move regex to module

cmlsharp:
A class that just contains static methods is essentially the same as a module so it seems kinda unidiomatic for `regex` to be a class. I get that it's a maybe a little weird to have a module with one function, but the whole reason for having regexes behind a namespace in the first place was for future-proofing.

Also, I noticed that `Missing` was not included in the import in __init__.py so I fixed that.

### DIFF
--- a/check50/__init__.py
+++ b/check50/__init__.py
@@ -38,16 +38,16 @@ from ._api import (
     exists,
     hash,
     include,
-    regex,
     run,
     log, _log,
     hidden,
-    Failure, Mismatch
+    Failure, Mismatch, Missing
 )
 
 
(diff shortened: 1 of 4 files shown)


Jelleas:
I'm not entirely sure this is worth a module of its own right now. Mostly because I can't foresee any other useful regexes just yet. I do agree we're better off using the `check50.regex.` namespace straight away, hence the previous implementation with a "class". 

I don't feel too strongly about this tbh. I agree a module is definitely more idiomatic and the right way long-term, and honestly that's probably a good enough reason to switch to that instead.

PR #250, opened 2021-03-08
temporarily disable logger writer

kzidane:


kzidane:
@brianyu28  and I looked a bit into this. When the CS50 library makes a `Logger.debug()` call (or any student code writes anything I imagine), that ends up calling `LoggerWriter.write()`, presumably because of how the output streams are being temporarily redirected when we enter the `CheckRunner` context. `LoggerWriter.write()` calls `self.logger.log()`, which writes to the `"check50"` logger, which we think tries to write to `stdout` or `stderr`, which causes `LoggerWriter.write()` to be called again and so on.

I'm setting the log level to `NOTSET` for now until we figure out if that's what's happening and what should happen instead.

PR #254, opened 2021-04-06
docs: fix broken json example.

pazz:
The extra comma violates the json spec and means that the content is not readable in python (see https://github.com/pazz/grade50/issues/1).

Also see

```
jid < caesar.json
invalid json format: invalid character ']' looking for beginning of value
```


dmalan:
Thanks!

PR #342, opened 2024-07-22
Windows Support for `check50`

Sashank999:
Refer to my PR cs50/lib50#79 for more context.

Summary:
1. Use `pexpect.PopenSpawn` instead of `pexpect.spawn`, for both `check50` and testsuite.
2. Use `threading.Timer` for raising timeouts in checks.
3. Add additional `exit()` when compiling checks.
    - This ensures that the check subprocesses run until their cleanup.
    - Existing pre-compiled checks are also modified.
4. Add `check50.run.isalive()` for checking subprocess status.
5. Add quoting for command line arguments.
6. Kill check subprocesses when timeout or EOF is raised.
7. Wait until the compiler's process exits, with a timeout of 30 seconds.

### DIFF
--- a/check50/__main__.py
+++ b/check50/__main__.py
@@ -345,6 +345,12 @@ def main():
         with lib50.ProgressBar("Checking") if "ansi" in args.output else nullcontext():
             # If developing, assume slug is a path to check_dir
             if args.dev:
+                # On Windows, the path is sometimes quoted. Resolving it as it is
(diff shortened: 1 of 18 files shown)


rongxin-liu:
Again thanks for your contribution, but we decided not to move forward with merging this PR per https://github.com/cs50/lib50/pull/79.

Issue #287, opened 2022-09-14
`check50` ANSI and HTML output do not make clear when `stdout` has unprintable characters


dmalan:
As in a bug like

```c
#include <stdio.h>

int foo(char c)
{
    printf('%c', c + 1);
    return 0;
}

int main(void)
{
    char c = 'a';
    printf('%c', foo(c));
}
```

which ends up printing `b` followed by an invisible NUL (because main` prints `foo`'s return value) followed by `\n`, which wouldn't match an expectation of `b\n`, but whose NUL student could not see in `check50`'s output.

goyalyashpal:
```diff
- which ends up printing `b` followed by an invisible NUL (because main`prints`foo`'s return value) followed by `\n`, which wouldn't match an expectation of `b\n`, but whose NUL student could not see in `check50`'s output.
+ which ends up printing `b` followed by an invisible NUL (because ``main`prints`foo``'s return value) followed by `\n`, which wouldn't match an expectation of `b\n`, but whose NUL student could not see in `check50`'s output.
```

Original in description: 

> which ends up printing `b` followed by an invisible NUL (because main`prints`foo`'s return value) followed by `\n`, which wouldn't match an expectation of `b\n`, but whose NUL student could not see in `check50`'s output.

I think what was intended:

> which ends up printing `b` followed by an invisible NUL (because ``main`prints`foo``'s return value) followed by `\n`, which wouldn't match an expectation of `b\n`, but whose NUL student could not see in `check50`'s output.

ivanharvard:
Fixed as of #356!

Tested using this snippet:
```c
#include <stdio.h>

int foo(char c)
{
    printf("%c", c + 1);
    return 0;
}

int main(void)
{
    char c = 'a';
    printf("%c\n", foo(c));
}
```

<img width="2169" height="560" alt="Image" src="https://github.com/user-attachments/assets/ce993506-b7ad-4724-8ff1-8ff619f7bef4" />

Issue #220, opened 2020-06-26
Invisible ASCII chars


Jelleas:
A common bug in Caesar and Vigenere is printing out characters that are not visible on screen. The most common cause here is actually through an off by one error, where the student loops to `< n+1` or `<= n` and then mistakenly encrypts the trailing `\0` to something exotic. However, check50s feedback isn't very helpful here:

![image](https://user-images.githubusercontent.com/1257775/85850892-fe2feb80-b7ad-11ea-8a50-ab88f339c205.png)

The above is the result of introducing this off-by-one in vault50's solution, full results at: https://submit.cs50.io/check50/db65ac87b1538ac1afc860e7df823b5158885d90

Can we improve?

tommylenz:
I had the very same issue when doing this pset myself! And I'd be happy to help here.
How about showing the ASCII chars that aren't visible usually in a different color in the actual output box?
Like it would say: "Actual Output: `ciphertext: b\0`", with the `\0 `in red, because it shouldn't be there, or yellow, because it's a char that's on our "seems-wrong-here" blacklist for all chars, that shouldn't ever be used in an output? Because chars like `\n` should be used sometimes 🤔

artemlazaruk:
I'm not sure that it's the same case, but I also have similar issues with Substitution problem and check50. 
https://submit.cs50.io/check50/78eb10b2c314dbe8fb51c41d475dd23b15a25e83

There are no errors or additional characters

```
#include <stdio.h>
#include <string.h>
#include <ctype.h>

int main(int argc, char *argv[])
{
    if (argc == 2)
    {
        char key[2][26];
        char alphabet[26] = "ABCDEFGHIJKLMNOPQRSTUVWXYZ";
        int n = strlen(argv[1]);
        int alpha = 0;
        int repeat = 0;

        // Assign key and alphabet to one array
        for (int i = 0; i < n; i++)
        {
            key[0][i] = alphabet[i];
            key[1][i] = toupper(argv[1][i]);

            // Counters for alphabetic characters and repeated characters
            if (isalpha(argv[1][i]) > 0)
            {
                alpha++;
            }

            for (int j = 0; j < n; j++)
            {
                if (argv[1][i] == argv[1][j])
                {
                    repeat++;
                }
            }
        }

        // Get plain text
        if (n == 26 && alpha == 26 && repeat == 26)
        {
            char plaintext[100];
            char ciphertext[100];

            printf("plaintext: ");
            scanf("%[^\n]s", plaintext);

[...]

PR #356, opened 2025-07-14
Added feature where invisible chars are highlighted in the Expected vs Actual output terminals

ivanharvard:
Fixes a UI issue in which Expected vs Actual outputs would appear identical, yet check50 would fail because there was a certain invisible character within the text. 

Related to #287 and #220.

+Added highlighting and escaping of newlines and spaces when they are the last character at the end of the output.
+Added a tooltip that describes the meaning of the highlighting.
+Added highlighting and escaping of arbitary invisible characters, mostly determined by Python's built-in `.isprintable()` function.

Color highlighting is gray by default, but can be easily modified by changing the color parameter when rendering the template.

image

### DIFF
--- a/check50/renderer/_renderers.py
+++ b/check50/renderer/_renderers.py
@@ -45,7 +45,7 @@ def _fmt_special_chars(txt, color):
         is_last = i == len(txt) - 1
 
         if not char.isprintable() and char not in ws_to_html:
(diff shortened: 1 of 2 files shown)


rongxin-liu (approved):
Looking good! Thanks! @ivanharvard

PR #357, opened 2025-07-15
add motivational msgs

### LINKED ISSUE #192: Introduce positive messages
We have happy, sad and neutral smileys, but we can do better to motivate and support students. Even simple messages like:

Good job!
You can do it!
Well done!

We can print messages like this (or a gif in the webview) every n runs, when we find that more tests pass than before, or perhaps when progress stalls.

bxngyn:
Fixes #192


Jelleas:
🎉Thanks for picking this up!

Small note from my side, be mindful of unintentional sarcasm. For instance if all checks suddenly fail (compile error), and the dice happens to land on `"~~~~~ You're getting there! ~~~~~"`

bxngyn:
Good point haha, I'll see what I can do to fix

rongxin-liu (approved):

rongxin-liu:
Looks good, merging!

PR #366, opened 2025-07-23
Simplify type handling in _truncate function

### LINKED ISSUE #365: check50.Mismatch from v4.0.0-dev fails on non-string inputs
Similar behavior as #185. Got fixed at the time but now is bugged again. Probably just need to attempt to convert the arguments in question to strings before we try and truncate.

Image

rongxin-liu:
Closes #365

### DIFF
--- a/check50/_api.py
+++ b/check50/_api.py
@@ -456,7 +456,7 @@ class Mismatch(Failure):
 
     def __init__(self, expected, actual, help=None):
         def _safe_truncate(x, y):
-            return _truncate(x, y) if x != EOF and x != TIMEOUT else x
+            return _truncate(x, y) if x not in (EOF, TIMEOUT) else x
 
         expected, actual = _safe_truncate(expected, actual), _safe_truncate(actual, expected)
 
--- a/check50/_api.py
+++ b/check50/_api.py
@@ -464,12 +464,6 @@ class Mismatch(Failure):
 
         super().__init__(rationale=rationale, help=help)
 
-        if expected == EOF:
-            expected = "EOF"
-
-        if actual == EOF:
-            actual = "EOF"
-
         self.payload.update({"expected": expected, "actual": actual})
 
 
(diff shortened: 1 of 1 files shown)


rongxin-liu:
Should also add test cases for the _truncate function.

ivanharvard:
I believe this may have a bug since EOF and TIMEOUT will be converted to strings, when they have no string representation, so you'll get something like `expected: "<class 'pexpect.exceptions.EOF'>"` instead of `expected: EOF`. You may have to convert them to strings before truncation?

Additionally, do we care about whether the quotes are included in the output of values? For instance, should raising `check50.Mismatch(10, 20)` result in `expected: "10"` or `expected: 10`?

ivanharvard:
Just pushed a commit that addressed the first concern. The second concern comes from `_raw()` manually adding those double quotes. Since we converted `EOF` to a string, `_raw()` will now output `"EOF"` instead of `EOF`.

```py
def _raw(s):
    """Get raw representation of s."""

    if isinstance(s, list):
        s = "\n".join(_raw(item) for item in s)

    if s == EOF or s == "EOF":
        return "EOF"
    elif s == TIMEOUT or s == "TIMEOUT":
        return "TIMEOUT"

    s = f'"{repr(str(s))[1:-1]}"'
    return s
```

This might be a fix, but then if a student were to literally print out `"EOF"`, and it somehow got passed through the `_raw()` function, then check50 may incorrectly display `"EOF"` as `EOF`. Any ideas?

ivanharvard:
Fixed the quotes around EOF and TIMEOUT.

bxngyn (commented):
LGTM

rongxin-liu:
> `check50.Mismatch(10, 20)` result in `expected: "10"` or `expected: 10`?

Good question! In this case, I’d prefer `expected: 10`. If people use check50.Mismatch("10", "20"), then the output should have quotes around it—`expected: "10"`—to indicate it is a string value.

@ivanharvard @bxngyn Let’s discuss it further, but as far as this PR is concerned, we’re good to merge.

Issue #185, opened 2019-10-09
Unexpected behaviour of check50.Mismatch for non-string arguments


mjdv:
When trying to raise Mismatch exceptions for non-string arguments, the first and last character of the string representation of the value are truncated, occasionally resulting in very cryptic error messages. Here are some examples.

```
>>> import check50
>>> raise check50.Mismatch(True, False)
Traceback (most recent call last):
  File "<stdin>", line 1, in <module>
check50._api.Mismatch: expected "ru", not "als"
>>> raise check50.Mismatch([], "")
Traceback (most recent call last):
  File "<stdin>", line 1, in <module>
>>> raise check50.Mismatch(10, 20)
Traceback (most recent call last):
  File "<stdin>", line 1, in <module>
check50._api.Mismatch: expected "", not ""
```

If the Mismatch exception class *only* allows string arguments, I believe this should explicitly be documented.

Additionally, while testing I ran into the following (related?) issue: once a Mismatch object has been created (but not raised), raising another Mismatch object causes a different (presumably unintended) exception.

```
>>> import check50
>>> check50.Mismatch("foo", "bar")
Mismatch('foo', 'bar')
>>> raise check50.Mismatch("baz", "lud")
Traceback (most recent call last):
  File "<stdin>", line 1, in <module>
  File "/home/mees/.local/lib/python3.6/site-packages/check50/_api.py", line 405, in __init__
    super().__init__(rationale=_("expected {}, not {}").format(_raw(expected), _raw(actual)), help=help)
TypeError: 'Mismatch' object is not callable
```

crossroads1112:
Problem 1 is fixed by the above commit. 

Problem 2 is actually pretty funny and threw me for a loop when I saw it.  For internationalization, we use python's `gettext` module which adds a function called `_` to the global namespace. Then any time you want a string to be translated based on the locale, you just call `_` on it. This is why you see `_("expected {}, not {}")` in the traceback. However, when the python interpreter runs in interactive mode, it defines a variable called `_` in the global namespace which is equal to the last evaluated expression. You can see this in action by opening up the interpreter in interactive mode and running

```
>>> 1 + 2
3
>>> _
3
>>> 5 + 6
11
>>> _ 
11
```

As such, in interactive mode if you type any expression at all after you `import check50`, any of the functions in `check50` that create internationalized output (i.e. most of them) will fail. You can even do something like:

```
>>> import check50
>>> def underscore_replacement(s): print("HELLO"); return s
[...]

PR #367, opened 2025-07-23
check50 will now warn users who try to import the library in a Python interactive mode

ivanharvard:
#185 has this quote:
> For internationalization, we use python's `gettext` module which adds a function called `_` to the global namespace. Then any time you want a string to be translated based on the locale, you just call `_` on it. This is why you see `_("expected {}, not {}")` in the traceback. However, when the python interpreter runs in interactive mode, it defines a variable called `_` in the global namespace which is equal to the last evaluated expression. 
> [...]
> Essentially, the interpreter overwrites the value of `_` set by `gettext` (i.e. the translation function) with whatever the value of the last expression you evaluated was (in your case the previous check50.Mismatch). So if you type any expression at all after you `import check50`, any of the functions in `check50` that create internationalized output (i.e. most of them) will fail with a similar error.

Since `check50` is practically never used in interactive mode (and wasn't primarily intended to), importing `check50` into interactive mode will now raise a warning.
image

### DIFF
--- a/check50/__init__.py
+++ b/check50/__init__.py
@@ -27,7 +27,7 @@ _setup_translation()
 import sys
 if hasattr(sys, 'ps1') or sys.flags.interactive:
     import warnings
-    warnings.warn(_("check50 is not intended for use in interactive mode."
+    warnings.warn(_("check50 is not intended for use in interactive mode. "
                     "Some behavior may not function as expected."))
 
 from ._api import (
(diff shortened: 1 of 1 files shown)


bxngyn (commented):
LGTM

rongxin-liu (approved):
Looks good. Thank you!

PR #370, opened 2025-08-04
--assertion-rewrite as an optional flag to override ENABLE_CHECK50_ASSERT

ivanharvard:
Previously, the only way to enable/disable assertion rewriting was by including this line at the top of your file:
```py
# ENABLE_CHECK50_ASSERT = 1
```
However, if the user now wishes to enable it without including this flag, they can simply run:
```
check50 . --assertion-rewrite
```
Otherwise, if they wish to override the assertion rewriting (even if `ENABLE_CHECK50_ASSERT` has been set to `1` or `True`), they can run:
```
check50 . --assertion-rewrite disabled
```

If both the argument passed in the command line and the comment left in the check file are conflicting, the presence of the flag takes precedence.

### DIFF
--- a/check50/__main__.py
+++ b/check50/__main__.py
@@ -334,6 +334,12 @@ def main():
     parser.add_argument("--no-install-dependencies",
                         action="store_true",
                         help=_("do not install dependencies (only works with --local)"))
+    parser.add_argument("--assertion-rewrite",
+                        action="store",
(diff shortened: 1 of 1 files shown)


rongxin-liu (approved):

rongxin-liu:
Looks good!

Issue #94, opened 2018-05-08
Codec failure


rbowden91:
Currently, if I run on the attached caesar.c:

check50 -d --offline cs50/2017/fall/caesar caesar.c

I get:

```
Error in sys.excepthook:
Traceback (most recent call last):
  File "/usr/local/lib/python2.7/dist-packages/check50.py", line 185, in excepthook
    cleanup()
  File "/usr/local/lib/python2.7/dist-packages/check50.py", line 170, in cleanup
    shutil.rmtree(config.tempdir)
  File "/usr/lib/python2.7/shutil.py", line 239, in rmtree
    onerror(os.listdir, path, sys.exc_info())
  File "/usr/lib/python2.7/shutil.py", line 237, in rmtree
    names = os.listdir(path)
OSError: [Errno 2] No such file or directory: '/tmp/tmpedXoxi'

Original exception was:
Traceback (most recent call last):
  File "/usr/local/bin/check50", line 9, in <module>
    load_entry_point('check50==2.2.2', 'console_scripts', 'check50')()
  File "/usr/local/lib/python2.7/dist-packages/check50.py", line 151, in main
    print_json(results)
  File "/usr/local/lib/python2.7/dist-packages/check50.py", line 250, in print_json
    print(json.dumps(output, cls=Encoder))
  File "/usr/lib/python2.7/json/__init__.py", line 251, in dumps
    sort_keys=sort_keys, **kw).encode(obj)
  File "/usr/lib/python2.7/json/encoder.py", line 207, in encode
    chunks = self.iterencode(o, _one_shot=True)
  File "/usr/lib/python2.7/json/encoder.py", line 270, in iterencode
    return _iterencode(o, 0)
UnicodeDecodeError: 'utf8' codec can't decode byte 0xb4 in position 13: invalid start byte

```

My current fix is appending ".decode('latin-1')" to the end of this line:
https://github.com/cs50/check50/blob/c054d44d74d6074c23b2ad6747fc14e2e14c3835/check50.py#L244

The issue is that the incorrect student submissions shifts an ascii value outside of the 0-127 range. Not sure if my fix is the right thing to do, but works for the time being.

rbowden91:
Oops, forgot to attach the file! (Changed from caesar.c to caesar.txt to appease GitHub)

[caesar.txt](https://github.com/cs50/check50/files/1986344/caesar.txt)

crossroads1112:
[...]

PR #373, opened 2025-08-05
Add authentication method flags

ivanharvard:
Allows user to run
```
check50 cs50/problems/2025/x/finance --ssh
```
or
```
check50 cs50/problems/2025/x/finance --https
```
which will change the `auth_method` argument in `push`, which was added in [PR #94 of lib50](https://github.com/cs50/lib50/pull/94).

If neither are included or both are included, the default behavior will occur (with a warning message appearing in the latter).

Using either one of these will change the progress bar's output as well:
image
image

### DIFF
--- a/check50/__main__.py
+++ b/check50/__main__.py
@@ -259,6 +259,20 @@ def process_args(args):
     if args.ansi_log and "ansi" not in seen_output:
         LOGGER.warning(_("--ansi-log has no effect when ansi is not among the output formats"))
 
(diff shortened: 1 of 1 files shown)


ivanharvard:
Added a more user-friendly message when pushing fails.
image

Also added traceback support.
image

rongxin-liu (approved):

rongxin-liu:
Looks good, merging! Thank you! @ivanharvard

PR #378, opened 2025-08-07
Added _process_list as a _raw and _truncate helper

### LINKED ISSUE #376: _raw(s) and _truncate(s, other) handle (nested) lists differently
In `_api.py`, `_raw` flattens nested lists while `_truncate` does not.

I believe it would be best to not flatten a nested list. A flattened list would obscure issues like these:
```
expected: 1
2
3
4
5
6
actual: 1
2
[3, 4, 5]
6
```
(although, in `Mismatch`, `_truncate` is called before `_raw`, so `Mismatch` will never truly flatten a nested list).

Or perhaps, `_raw` should not apply new lines to lists at all, considering we want the raw representation of the object in string form.

ivanharvard:
`_process_list` now handles converting lists to strings. `_raw` will no longer handle "raw"-ing a list, as that will be outsourced to `_process_list`. This closes #376, since the responsibilities of each function have now been changed.

### DIFF
--- a/check50/_api.py
+++ b/check50/_api.py
@@ -425,6 +425,8 @@ class Missing(Failure):
     """
 
     def __init__(self, missing_item, collection, help=None):
+        if isinstance(collection, list):
+            collection = _process_list(collection, _raw)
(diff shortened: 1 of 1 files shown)


rongxin-liu (approved):
Looking good! Thank you!

PR #385, opened 2025-09-03
Rm attrs

Jelleas:
Per @cmlsharp's comment in https://github.com/cs50/check50/pull/368, Python 3.7 brought with it dataclasses that can directly replace `attrs`. This pr replaces the only uses of `attrs` in `check50` with dataclasses and removes `attrs` as a dependency. 

Opted to use python 3.10 as the 4.0.0-dev branch is already using features from Python 3.10. Such as `match`, and   `importlib.resources.files` that was added in Python 3.9.

Used python 3.10 syntax for type hints (notably the use of `|`, and `list` instead of `typing.List`).
Set `slots=True` which is only supported from 3.10 onwards.

### DIFF
--- a/check50/__main__.py
+++ b/check50/__main__.py
@@ -1,5 +1,6 @@
 import argparse
 import contextlib
+import dataclasses
 import enum
 import gettext
 import importlib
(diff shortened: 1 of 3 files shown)


rongxin-liu:
Thank you! @Jelleas

PR #392, opened 2025-11-13
Create mahdiyar.py

mahdiyar23-btc:
def convert(text):
    text = text.replace("(:" , "😃")
    text = text.replace("):" , "😞")
    return text
def main():
    user_input = input("what's emoji?")
    convert_text = convert(user_input)
    print(convert_text)
    main()


rongxin-liu:
Thank you for your contribution, but I don’t think this adds much to check50.

